PR #63, opened 2021-08-29
Remove service links

ableuler:
This PR removes the service IPs from the user pod envs.


rokroskar (approved):
Nice - I forgot this was solved already in renku-notebooks :) https://github.com/SwissDataScienceCenter/renku-notebooks/pull/431

PR #64, opened 2021-08-29
Remove traefik

ableuler:
This PR removes traefik and used the oauth2 proxy as a proxy and not just as forwardAuth middleware for traefik. As a side-effect of this, we also remove the custom authorization plugin which used to check if the `userId` added to the request by the oauth2 proxy matches the userId given as part of the JupyterServer spec. Instead, we rely on the built-in oauth2 proxy options, which are specifying a list of allowed Emails (read from the "email" claim) and/or a list of allowed groups (read from the "groups" claim). The CRD thus changes as part of this PR.

### DIFF
--- a/tests/unit/test_manifest_auth.py
+++ b/tests/unit/test_manifest_auth.py
@@ -71,23 +71,15 @@ def test_auth_oidc(oidc_secret, valid_spec):
     secret = manifest["secret"]
     assert any(
         [
-            container["name"] == "authentication-plugin"
+            container["name"] == "oauth2-proxy"
             for container in manifest["statefulset"]["spec"]["template"]["spec"][
                 "containers"
             ]
         ]
     )
-    assert any(
-        [
-            container["name"] == "authorization-plugin"
-            for container in manifest["statefulset"]["spec"]["template"]["spec"][
-                "containers"
-            ]
-        ]
-    )
-    assert "authenticationPluginCookieSecret" in secret["data"].keys()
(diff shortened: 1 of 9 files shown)


tests/unit/test_manifest_auth.py
```diff
@@ -71,23 +71,15 @@ def test_auth_oidc(oidc_secret, valid_spec):
...
-            for container in manifest["statefulset"]["spec"]["template"]["spec"][
-                "containers"
-            ]
-        ]
-    )
-    assert "authenticationPluginCookieSecret" in secret["data"].keys()
+    assert "oauth2ProxyCookieSecret" in secret["data"].keys()
     auth_container = manifest["statefulset"]["spec"]["template"]["spec"]["containers"][
-        3
+        1
```
rokroskar, line 82: is there really no other way to do this? :)

ableuler: The problem is that for the purpose of patching the resources resulting from the template rendering, we have to assume a certain order of the containers in the pod. In a first implementation - for exactly that reason - I actually treated each container as a separate resource which was addressable by name and only added them to the pods container array before applying the statefulset to the cluster. However, this was really confusing, because you're patching the proper kubernetes resources everywhere else. This is definitely something that has to be documented better.
So failing tests if the order of containers in the pod is wrong is actually a good thing.

ableuler: Your comment actually made me double check: 
https://github.com/json-patch/json-patch2/issues/18

rokroskar (approved):

PR #68, opened 2021-09-09
feat: idle sessions culling

olevski:
Andreas you were right about the single quotes in the `--skip-auth-path` option for the oauth2-proxy. After removing them things work as expected.

The metrics from the oauth2-proxy are not very reliable. This is because the number of requests increases sometimes when the user is not really doing anything (as far as I can tell). So I fully removed this. Using the information from the jupyter server is the best from the tests that I did. To be safe I also added the total cpu usage for the pod (if the metrics are available). 

The basic functionality is as follows:
- every X seconds (based on what the user selects) check the cpu usage and information from jupyter server /api/status endpoint
- based on the data from the previous step determine if the session is idle
- if the session is idle then mark it so with an annotation - assume that the sessions has been idle for a duration equal to the interval at which the idle check runs
- if the session has been idle for more than a specific threshold (based on the annotation that keeps track of the idle duration) then delete it
- if the session is not in an idle state anymore then reset the idle timer in the annotation


olevski:
@ableuler the only way I got get the probes to work is by using an `exec` probe. Otherwise the startup probe kept failing. Also by using exec probes I avoid problems and weird interactions where the service is disabled with the liveness probe but we need the service in order to do the liveness and all other probes because we need to be routed through the proxy.

olevski:
Also this is currently deployed at tasko.dev.renku.ch.

controller/templates/statefulset.yaml
```diff
@@ -103,32 +103,38 @@ spec:
...
           livenessProbe:
-            httpGet:
-              httpHeaders:
-                - name: host
-                  value: 127.0.0.1:8888
-              path: {{ path }}
-              port: 8888
+            exec:
+              command:
+                - "curl"
```
ableuler, line 108: I thought about this too, but this requires curl to be installed inside the jupyter session, which is actually not always the case (for example `jupyter/minimal-notebook` doesn't have it).

helm-chart/amalthea/templates/deployment.yaml
```diff
@@ -66,6 +66,12 @@ spec:
                 {{- include "amalthea.selectorLabels" . | nindent 16 }}
             - name: EXTRA_CHILD_RESOURCES
               value: {{ .Values.extraChildResources | toJson | quote }}
+            - name: SESSION_IDLE_CHECK_INTERVAL_SECONDS
```
ableuler, line 69: This is a bit picky, but SESSION is somewhat of a Renku terminology, so I'd either omit `SESSION_` it completely or use `JUPYTER_SERVER_` instead.

helm-chart/amalthea/values.yaml
```diff
@@ -126,3 +126,14 @@ crdNames:
...
   shortNames:
     - js
+
+
+# Determine how often amalthea checks each session to determine
+# whether it is idle and how long should one session be idle for
+# before it is culled. The total cpu usage of all containers in a 
+# user session's pod is used to determine if the session is idle
+# in addition to the open connections to the oauth2-proxy.
+culling:
```
ableuler, line 136: I think the default should be no culling. I can see many use-cases where this would be the desired behaviour. So either a `culling.enabled` flag or the option to omit culling by setting all numeric values to 0 would be nice.

controller/server_controller.py
```diff
@@ -125,6 +127,101 @@ def create_fn(labels, logger, name, namespace, spec, uid, **_):
...
+    config.custom_resource_name,
+    interval=config.SESSION_IDLE_CHECK_INTERVAL_SECONDS,
+)
+def cull_idle_sessions(body, name, namespace, logger, **kwargs):
+    """
+    Check if a session is idle (has zero open connections in proxy and CPU is below
+    threshold). If the session is idle then annotate the jupyterserver resource with
+    the idle duration. If any sessions have been idle for long enough, then cull them.
+    """
+    try:
```
ableuler, line 142: In some cases (for example when having a long-running init container) - there is no main pod status. I'd check for this first and return right away, something like |
```
if not "mainPod" in body["status"]: 
   return
```

controller/server_controller.py
```diff
@@ -125,6 +127,101 @@ def create_fn(labels, logger, name, namespace, spec, uid, **_):
...
+        pod_name = body["status"]["mainPod"]["name"]
+    except KeyError:
+        pod_name = None
+    cpu_usage = get_cpu_usage(pod=pod_name, namespace=namespace)
+    js_server_status = get_js_server_status(body)
+    custom_resource_api = get_api(config.api_version, config.custom_resource_name)
+    idle_seconds = int(
```
ableuler, line 149: I think it would be a bit nicer to write this into the status of the custom resource into the body["status"] where all the other status information about the child resources is stored too.

ableuler (changes requested):

ableuler:
Thanks for this and also for trying to fix the probe. I suggest to just remove the probe for the time being (you can make a separate commit for this as it's clearly broken) and add the culling through this PR, we can handle the probe later.

controller/culling.py
```diff
@@ -0,0 +1,85 @@
...
+    except ApiException as err:
+        logging.warning(
+            f"Could not get CPU usage for culling idle sessions for pod {pod}, because: {err}"
+        )
+        return total_usage_milicores
+
+    try:
+        for container in res.containers:
+            if container["usage"]["cpu"].endswith("n"):
+                total_usage_milicores += int(container["usage"]["cpu"][:-1]) / 1e6
```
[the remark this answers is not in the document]
olevski: This seems to confirm we get nanocores from the metrics server: https://discuss.kubernetes.io/t/metric-server-cpu-and-memory-units/7497. 

A few other articles / forums mention this but could not find any official documentation.

I did try to run a pod that was consuming 3-4 cores and still got nanocores. So it seems it is always using nanocores.

controller/templates/statefulset.yaml
olevski: Ok yeah that explains the failing tests. I will revert the changes to the probes and turn them back off.

controller/config.py
```diff
@@ -44,3 +44,13 @@
...
     else int(os.getenv("KOPF_CREATE_RETRIES"))
 )
+
+SESSION_IDLE_CHECK_INTERVAL_SECONDS = int(
+    os.getenv("SESSION_IDLE_CHECK_INTERVAL_SECONDS", 300)
+)
+SESSION_CULL_IDLE_THRESHOLD_SECONDS = int(
+    os.getenv("SESSION_CULL_IDLE_THRESHOLD_SECONDS", 86400)
+)
+CPU_USAGE_MILICORES_IDLE_THRESHOLD = int(
```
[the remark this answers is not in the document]
olevski: 🤦  you are right it is with a double "l"

controller/server_controller.py
```diff
@@ -125,6 +127,88 @@ def create_fn(labels, logger, name, namespace, spec, uid, **_):
...
+        ).total_seconds()
+        > config.JUPYTER_SERVER_IDLE_CHECK_INTERVAL_SECONDS
+    )
+
+    if (
+        jupyter_server_is_idle
+        and idle_seconds + config.JUPYTER_SERVER_IDLE_CHECK_INTERVAL_SECONDS
+        >= config.JUPYTER_SERVER_CULL_IDLE_THRESHOLD_SECONDS
+    ):
+        logger.info(f"Deleting session {name} due to inactivity")
```
ableuler, line 168: Maybe add the namespace here in case one amalthea instance is looking after servers in different namespaces.

ableuler:
```suggestion
        logger.info(f"Deleting jupyter server {name} due to inactivity")
```

controller/server_controller.py
```diff
@@ -125,6 +127,88 @@ def create_fn(labels, logger, name, namespace, spec, uid, **_):
...
+        try:
+            custom_resource_api.delete(name=name, namespace=namespace)
+        except NotFoundError:
```
ableuler, line 171: I think this is a bit generous and has the potential of hiding actual bugs. Catching the `NotFoundError` makes sense to me, but I'd at least log a warning along the lines of: "Trying to delete jupyter server X in namespace Y, but we can not find it. Has it been deleted in the mean time?")

ableuler: Same applies for the patch requests further down.

controller/server_controller.py
```diff
@@ -125,6 +127,88 @@ def create_fn(labels, logger, name, namespace, spec, uid, **_):
...
+            pass
+        return
+
+    if jupyter_server_is_idle:
+        logger.info(
+            f"Session {name} found to be idle for "
```
ableuler, line 177:
```suggestion
            f"Jupyter Server {name} in namespace {namespace} found to be idle for "
```

controller/server_controller.py
```diff
@@ -125,6 +127,88 @@ def create_fn(labels, logger, name, namespace, spec, uid, **_):
...
+                    },
+                },
+                content_type=CONTENT_TYPES["merge-patch"],
+            )
+        except NotFoundError:
+            pass
+    else:
+        if idle_seconds > 0:
+            try:
+                logger.info(f"Resetting idle timer for session {name}.")
```
ableuler, line 199:
```suggestion
                logger.info(f"Resetting idle timer for Jupyter server {name} in namespace {namespace}.")
```

helm-chart/amalthea/templates/rbac/_rbac_rules.tpl
```diff
@@ -7,12 +7,17 @@
...
   - apiGroups: [{{ .Values.crdApiGroup }}]
     resources: [{{ .Values.crdNames.plural }}]
-    verbs: [list, watch, patch]
+    verbs: [list, watch, patch, delete]
 
   - apiGroups: [""]
     resources: [pods]
     verbs: [get, list, watch, delete]
 
+  # Amalthea get pod metrics used to cull idle sessions
```
ableuler, line 16:
```suggestion
  # Amalthea get pod metrics used to cull idle Jupyter servers
```

ableuler (changes requested):

ableuler (approved):
Let's merge this! Also, I think you can tag a rc10 release here!

PR #77, opened 2021-09-16
docs: automatically generate CRD from helm template

ableuler:
The intent of this PR is to have our crd automatically turned into docs using https://doc.crds.dev/

Unfortunately, it doesn't work yet: https://github.com/crdsdev/doc/issues/156


ableuler:
closing as we have an ok alternative and the upstream issue might never get fixed.

PR #80, opened 2021-09-20
fix(chart): enable http probes for jupyter server

olevski:
Enables http probes for the Jupyter server container.

Deployed on tasko.dev.renku.ch

Tested to make sure:
- streamlit works in both registered and anonymous sessions
- jupyter server endpoint (8888) cannot be accessed from another jupyter server
- jupyter server endpoint (8888) cannot be accessed from a random pod in the cluster
- culling still functions as expected

Project to test streamlit: https://tasko.dev.renku.ch/projects/tasko.olevski/test-streamlit

closes https://github.com/SwissDataScienceCenter/renku-notebooks/issues/741


ableuler (changes requested):
This does what it promises, only minor changes suggested. Seeing it code in front of my eyes I'm a bit worried though that the approach of making the oauth2 proxy a non-auth2-proxy through the config might be a bit error prone. I'd suggest the following:
- we merge this PR the way it is now (modulo the minor suggestions)
- we add tests to this repo that test the oauth2 flow (already did some experiments with dex). when doing this, we also make sure that we test a scenario where access is _not_ permitted.
- maybe it would make sense to replace the oauth2 proxy with something like http://tinyproxy.github.io/ in the non-oidc case? This way the config for the jupyter server would stay exactly the same, but we'd swap out the proxy container?

controller/templates/statefulset.yaml
```diff
@@ -104,30 +96,31 @@ spec:
...
+          startupProbe:
+            httpGet:
+              path: {{ path }}
+              port: 8888
+            periodSeconds: 10
+            failureThreshold: 30
+            timeoutSeconds: 9
+        
         - name: oauth2-proxy
           image: "bitnami/oauth2-proxy:7.1.3"
```
aledegano, line 122: I know this is not the related to this PR, but I'm noticing now: the official Docker image of oauth2-proxy is stored in quay.io: https://quay.io/oauth2-proxy/oauth2-proxy maybe we should consider using that image? As it would give us some more guarantee of the contents?

ableuler: I like about the Bitnami images that they usually do not run as root. Other than that I can not think of any reason to not use the "official" image. Feel free to open an issue for it.

controller/templates/secret.yaml
```diff
@@ -7,10 +7,10 @@ data:
...
   jupyterServerCookieSecret: {{ jupyter_server_cookie_secret | b64encode }}
 
-  {% if auth["oidc"]["enabled"] %}
-  {% if "value" in auth["oidc"]["clientSecret"] %}
+  {% if auth["oidc"]["enabled"] and "value" in auth["oidc"]["clientSecret"] %}
   oidcClientSecret: {{ oidc["clientSecret"]["value"] | b64encode }}
-  {% endif %}
-  oauth2ProxyCookieSecret: {{ authentication_plugin_cookie_secret | b64encode }}
+  {% else %}
+  oidcClientSecret: {{ name | b64encode }}
```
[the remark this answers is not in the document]
olevski: @ableuler yes the proxy complains when some arguments are missing:
```
➜  renku docker run bitnami/oauth2-proxy:7.1.3
[2021/09/22 11:07:15] [main.go:54] invalid configuration:
  missing setting: cookie-secret
  provider missing setting: client-id
  missing setting: client-secret or client-secret-file
  missing setting for email validation: email-domain or authenticated-emails-file required.
      use email-domain=* to authorize all email addresses
```

olevski:
@ableuler I agree that the code for making the auth2proxy pass things through looks complicated and increases the risks for errors and accidentally making the oidc not work properly when we actually want it to.

Out of your suggestions I like the third one the most. I even tested a version of this PR with traefik where I set it up to just pass things through. So I think that using tinyproxy is the best thing to do. I will edit this PR to do that.

There is a bit more code this way but it is way easier to understand and maintain.

ableuler:
> @ableuler I agree that the code for making the auth2proxy pass things through looks complicated and increases the risks for errors and accidentally making the oidc not work properly when we actually want it to.
> 
> Out of your suggestions I like the third one the most. I even tested a version of this PR with traefik where I set it up to just pass things through. So I think that using tinyproxy is the best thing to do. I will edit this PR to do that.
> 
> There is a bit more code this way but it is way easier to understand and maintain.

ok, sounds good. AFAIK a mostly idle traefik instance has a memory footprint of ~60 MB, compared to the (advertised) 2MB of tinyproxy. So I think it's worth the effort of figuring out how to use it here.

olevski:
I wanted to document here why I did not end up using tiny proxy.

Basically the main reason is because tinyproxy is quite sensitive to the slashes and path you have defined in the config. For example with a config like this:

```
User nobody
Group nogroup
Port 4180
Listen 0.0.0.0
ReverseOnly Yes
ReverseMagic Yes
ReversePath "/sessions/anon-2di4w-test1-6deee93f" "http://localhost:8888/sessions/anon-2di4w-test1-6deee93f/"
ReverseBaseURL "https://tasko.dev.renku.ch/sessions/anon-2di4w-test1-6deee93f
```

The proxy will not recognize the url for the notebook like this:
`https://tasko.dev.renku.ch/sessions/anon-2di4w-test1-6deee93f?token=anon-I4wtWzj50y1_A5FVXlWvu3pG_zlOQCz-1wl3xV_NLwY`

But this will work:
`https://tasko.dev.renku.ch/sessions/anon-2di4w-test1-6deee93f/?token=anon-I4wtWzj50y1_A5FVXlWvu3pG_zlOQCz-1wl3xV_NLwY`

So it expects that extra slash before the token. And I kept looking for a way to include something like `.*` in the regex but it does not work. This sensitivity to the / is a bit scary and the lack of functionality to use something like `.*`. This is perhaps not a big deal now but it may be a problem for other programs that users may run (i.e. rstudio, etc) that need to pass things as query parameters but in this case tinyproxy will not accept them because they have not been specifically mentioned in the config.

controller/templates/statefulset.yaml
```diff
@@ -177,4 +179,26 @@ spec:
             limits:
               cpu: 100m
               memory: 32Mi
+        {% else %}
+        - name: passthrough-proxy
+          image: traefik:v2.5
+          args:
+           - --entryPoints.web.address=:4180
+           - --providers.file.directory=/traefik
+           - --log.level=DEBUG
```
ableuler, line 188: I think this should not be the default here as it makes for heavy logs and increases the risk of accidentally logging some credentials.

controller/templates/statefulset.yaml
```diff
@@ -177,4 +179,26 @@ spec:
...
+              containerPort: 4180
+              protocol: TCP
+          volumeMounts:
+            - name: traefik-dynamic-config
+              mountPath: /traefik/traefik-dynamic-config.yaml
+              subPath: traefik-dynamic-config.yaml
+          resources:
+            requests:
+              cpu: 20m
+              memory: 16Mi
```
ableuler, line 200:
```suggestion
              memory: 32Mi
```
I did some testing and it was very easy for me to push the memory consumption above 16Mi and almost impossible (actually I didn't manage) to go above 32Mi. So I think with 32 we should be on the pretty save side (which is better than I expected).

controller/templates/statefulset.yaml
ableuler:
```suggestion
```

ableuler (changes requested):
Two one-liner suggestions only. Sorry again for my changes of opinion on this issue, but I'm really happy with how it has turned out in the end. I think always having _a_ proxy inside the pod gives us a nice flexibility in separating whatever is happening inside and outside the of the pod and the overhead in both resources and configuration is manageable.

ableuler:
> I wanted to document here why I did not end up using tiny proxy.

Thanks for the writeup, good to have this documented

ableuler (approved):

PR #84, opened 2021-09-22
chore(tests): do not push docker images to renku repo when doing tests

olevski:
Should fix the failing workflows from PRs that are not part of the SDSC org. There is no need to login to the renku docker repo to push images for testing.


aledegano (dismissed):

ableuler (approved):
thx for the quick fix!

PR #88, opened 2021-09-22
(feat) Add scheduler to pack users sessions.

### LINKED ISSUE #76: Optimize user pod placement
This feature has been moved around a little bit: swissdatasciencecenter/renku-notebooks#570 and swissdatasciencecenter/terraform-renku#120, to finally land here.

We identified this project as the best place where to offer the option of scheduling user pod "packed" in the nodes, which in turns allows for an easier scaling of the user-dedicated nodes.

To achieve the above, Amalthea will (optionally) deploy a secondary scheduler in the Kubernetes cluster, which will have the policy `MostRequestedPriority` with highest weight and the user pods will be scheduled by this scheduler.

Docs: https://kubernetes.io/docs/reference/scheduling/policies/

aledegano:
The chart can deploy an additional scheduler, with arbitrary configuration, that can then be used to schedule the operator-managed pods according to the chosen strategy.

To test create a JupyterServer resource as:
```yaml
---
apiVersion: amalthea.dev/v1alpha1
kind: JupyterServer
metadata:
  name: test-spark-session
  namespace: datascience-workloads
spec:
  jupyterServer:
    image: jupyter/all-spark-notebook:latest
  routing:
    host: test.empty-adegano.iac.dev.renku.ch
    path: /spark-session
    tls:
      enabled: true
      secretName: test-empty-tls
    ingressAnnotations:
      cert-manager.io/cluster-issuer: letsencrypt-production
      kubernetes.io/ingress.class: nginx
  patches:
    - patch:
        - op: add
          path: /statefulset/spec/template/spec/schedulerName
          value: "datascience-workloads-scheduler"
      type: application/json-patch+json
```

Close #76.


helm-chart/amalthea/values.yaml
```diff
@@ -127,13 +128,47 @@ crdNames:
...
+      weight: 1
+    - name: LeastRequestedPriority
+      weight: 1
+    - name: BalancedResourceAllocation
+      weight: 1
+    - name: NodePreferAvoidPodsPriority
+      weight: 1
+    - name: NodeAffinityPriority
+      weight: 1
+    - name: TaintTolerationPriority
```
ableuler, line 173: Does that mean that our taints/tolerations with which we currently force Jupyter servers onto certain nodes would be overruled by the dense packing policy?

ableuler (commented):
Looking forward to having this feature. I have one request regarding the "interface" of it. When the additional scheduler is enabled through the helm chart, I would expect this scheduler to be added to the JupyterServers per default without the need for a patch. I think this could be done by passing a "SERVER_SCHEDULER_NAME" environments variable to the controller deployment spec from where it would then be added to the statefulset spec during or right after the rendering of the jinja template. What do you think?

helm-chart/amalthea/values.yaml
aledegano: No: they coexists, thus the scheduling policy only applies to the nodes that still satisfy the taints/tolerations/affinity declared by the deployments.

aledegano:
> Looking forward to having this feature. I have one request regarding the "interface" of it. When the additional scheduler is enabled through the helm chart, I would expect this scheduler to be added to the JupyterServers per default without the need for a patch. I think this could be done by passing a "SERVER_SCHEDULER_NAME" environments variable to the controller deployment spec from where it would then be added to the statefulset spec during or right after the rendering of the jinja template. What do you think?

Yeah, it's a good idea, I'll implement it!

helm-chart/amalthea/values.yaml
ableuler: 👍 perfect!

olevski:
@aledegano I see that you have renamed the `helm-chart/amalthea/values.schema.json`. It seems that this mean that the values file will not be checked against the schema anymore?

Can we re-enable this? I think that it is definitely ok to not validate the sections that are added to the values file for the scheduler. You can just add the top most level sections to the schema and allow them to contain anything in their more nested levels.

aledegano:
> @aledegano I see that you have renamed the `helm-chart/amalthea/values.schema.json`. It seems that this mean that the values file will not be checked against the schema anymore?
> 
> Can we re-enable this? I think that it is definitely ok to not validate the sections that are added to the values file for the scheduler. You can just add the top most level sections to the schema and allow them to contain anything in their more nested levels.



> @aledegano I see that you have renamed the `helm-chart/amalthea/values.schema.json`. It seems that this mean that the values file will not be checked against the schema anymore?
> 
> Can we re-enable this? I think that it is definitely ok to not validate the sections that are added to the values file for the scheduler. You can just add the top most level sections to the schema and allow them to contain anything in their more nested levels.

Oh, yes, sorry that was only temporary to try the values before having to write the validation. I'll rename it and add the new possible values.

aledegano:
@ableuler In the second commit I've modified the code to automatically schedule the Jupiterservers with the optional scheduler when that is enabled, let me know if you are okay with the way the value is passed along and set.

@olevski I've re-enabled the schema and added the part relevant to the scheduler.

helm-chart/amalthea/Chart.yaml
```diff
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.0.1-rc1
```
olevski, line 18: I would either put this back the way it was or change it to `0.1.0-rc11`. I will submit a PR to tag a proper release.

helm-chart/amalthea/values.yaml
```diff
@@ -76,10 +76,10 @@ extraChildResources: []
   #   group: amalthea.dev
 
 image:
-  repository: renku/amalthea
```
olevski, line 79: Reminder to not leave this in.

helm-chart/amalthea/values.yaml
```diff
@@ -76,10 +76,10 @@ extraChildResources: []
...
+  repository: degano/amalthea
   pullPolicy: IfNotPresent
   # Overrides the image tag whose default is the chart appVersion.
-  tag: "0.0.1-rc1"
+  tag: '0.1.0-rc11.n006.h46d1a71'
```
olevski, line 82: Same as the previous comment about the chart tag. I always forget and check this in with chartpress.

olevski (changes requested):
Super minor changes. That are easy to address. All else looks good to me.

controller/config.py
```diff
@@ -51,3 +51,4 @@
 CPU_USAGE_MILLICORES_IDLE_THRESHOLD = int(
     os.getenv("CPU_USAGE_MILLICORES_IDLE_THRESHOLD", 200)
 )
+SERVER_SCHEDULER_NAME = os.getenv("SERVER_SCHEDULER_NAME", "")
```
ableuler, line 54:
```suggestion
SERVER_SCHEDULER_NAME = os.getenv("SERVER_SCHEDULER_NAME", None)
```

controller/templates/statefulset.yaml
```diff
@@ -60,7 +60,9 @@ spec:
       securityContext:
         fsGroup: 100
       enableServiceLinks: false
-
+      {% if schedulerName|length %}
```
ableuler, line 63:
```suggestion
      {% if scheduler_name %}
```

controller/templates/statefulset.yaml
```diff
@@ -60,7 +60,9 @@ spec:
...
+      schedulerName: {{schedulerName}}
```
ableuler, line 64:
```suggestion
      schedulerName: {{ scheduler_name }}
```

controller/k8s_resources.py
```diff
@@ -68,6 +70,7 @@ def create_template_values(name, spec):
         "path": os.path.join("/", spec["routing"]["path"].rstrip("/")),
         "pvc": spec["storage"]["pvc"],
         "routing": spec["routing"],
+        "schedulerName": config.SERVER_SCHEDULER_NAME,
```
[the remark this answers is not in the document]
ableuler: Sorry for being picky here, just trying to keep python variables and json object keys a bit separate.

ableuler (changes requested):
Very nice. I left a few suggestions, mostly about code styling. It's mostly about taste and what you suggested definitely works, so feel free to discard or accept as you like :)

tests/unit/test_k8s_resources.py
```diff
@@ -75,6 +75,7 @@ def test_create_template_values(valid_spec):
         "path",
         "pvc",
         "routing",
+        "schedulerName",
```
ableuler, line 78:
```suggestion
        "scheduler_name",
```
according to my other suggestions

olevski (approved):

ableuler (approved):

PR #94, opened 2021-09-23
chore: prep 0.1.0 release

olevski:


helm-chart/amalthea/Chart.yaml
```diff
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.0.1-set.by.chartpress
+version: 0.1.0
```
ableuler, line 18: This will be correctly set by chartpress before publishing the chart. However, I think increasing this number is a good idea. But then we'd have to do it by setting `resetVersion` in `chartpress.yaml` to something like `0.1.0-dev` now and bump it there with every release. Just bumping the number here will not be long lived as chartpress will overwrite it with `0.0.1-set.by.chartpress` in the one action.

CHANGELOG.md
```diff
@@ -0,0 +1,55 @@
+# [0.1.0](https://github.com/SwissDataScienceCenter/amalthea/compare/da735ced323eacb38fd010e4ae0a0479fb2bf310...0.1.0) (2021-09-15)
+
+
+### Bug Fixes
```
ableuler, line 4: Thanks for adding the CHANGELOG file. Since this is the first entry here, I think a very short list of the most important features which are already included is more meaningful than the conventional changelog.

ableuler (changes requested):

helm-chart/amalthea/Chart.yaml
ableuler: See https://github.com/jupyterhub/chartpress#configuration for reference

ableuler (approved):
Nice summary - that actually sounds like a useful project! 😊

PR #99, opened 2021-09-24
chore: sort/beautify CRD add some missing descriptions

ableuler:
Note: I seperated the spec part of the CRD which is plain YAML from the helm template to make it easier for editors to auto-format the CRD.


aledegano (approved):
Great idea!

PR #104, opened 2021-10-04
feat(app): add resource usage to JupyterServer resources

### LINKED ISSUE #74: Add resource usage in jupyterserver status
This would enable the status of the server (in terms of cpu usage, memory, disk) to be tracked and displayed by renku notebooks / ui , etc.

This can be done by adding a kopf timer that will periodically update the resource.

olevski:
Periodically check and update the status of JupyterServer resources with:
- disk usage acquired from running the `df` linux command in any container that has the `workspace` volume mounted
- cpu and memory usage from the k8s metrics api

closes #74


olevski:
@ableuler I can try to see if I can query the metrics endpoint of the kubelet.

And btw we already rely on the metrics server (albeit partially) for culling. Also this is not a strong requirement because if the metrics server is not there then you can still run Amalthea without a problem but you will not get resource usage information.

olevski:
@ableuler I have not really had a chance to dig into this. As we discussed I think it is a good idea to merge this and then examine using the kubelet as you suggest. I opened an issue for this so that we do not lose track of it.

See #114

ableuler:
sounds good to me!

ableuler (approved):

PR #110, opened 2021-10-21
feat(app): option to not limit size of user session emptyDir

### LINKED ISSUE #105: Make user server EmptyDir size limits optional
When PVCs are not used for a user session, then the requested disk size is set as the `sizeLimit` parameter for an `emptyDir`. 

K8s then immediately evicts the user session when more than the allowed space is consumed by the session. Therefore to improve the UX it is a good idea to have the option (preferrably in the helm chart) that would allow someone to deploy Amalthea and not use PVCs but then also to not include the `sizeLimit` property so that user sessions are not evicted.

olevski:
closes #105

### DIFF
--- a/helm-chart/amalthea/crd-spec.yaml
+++ b/helm-chart/amalthea/crd-spec.yaml
@@ -216,7 +216,7 @@ spec:
               type: string
           type: object
         size:
-          default: null
+          default: 100Mi
           description:
             Size of the PVC or sizeLimit of the emptyDir volume which backs the
             session respectively.


helm-chart/amalthea/crd-spec.yaml
```diff
@@ -221,5 +221,14 @@ spec:
             Size of the PVC or sizeLimit of the emptyDir volume which backs the
             session respectively.
           type: string
+        emptyDir:
+          type: object
+          description: "Options related to using emptyDir for storage. If PVC storage is disabled then emptyDir storage is used, otherwise all options here are ignored."
+          default: {}
+          properties:
+            enforceSizeLimit:
```
ableuler, line 229: I'm wondering if we could just interpret 
```
pvc: 
  enabled: false
  size: {null | ""}
```
as setting for an `emptyDir` volume without any size limit. It's a bit shorter and - probably more importantly - we can do it without changing the CRD.

controller/templates/pvc.yaml
```diff
@@ -7,7 +7,11 @@ spec:
     - ReadWriteOnce
   resources:
     requests:
+      {% if storage["size"] is not none and storage["size"] != "" %}
```
ableuler, line 10: oh, I didn't know that Jinja uses `is not none` instead of python style `is not None`! 👍

helm-chart/amalthea/crd-spec.yaml
```diff
@@ -216,7 +216,7 @@ spec:
               type: string
           type: object
         size:
-          default: 100Mi
+          default: null
```
ableuler, line 219: I'd leave the default for now and just set `null` explicitly when needed, for example in renku-notebooks. We can start collecting a list of things that we want to change when we update the CRD to `v1beta1` or `v1alpha2` (and I agree that changing this default could be one of them), but I think it would be nice to defer this for a bit longer and keep the CRD stable.

ableuler:
```suggestion
          default: 100Mi
```

PR #112, opened 2021-11-04
fix(app): faililng k8s probes for rstudio

olevski:
So it seems that rstudio returns 404 even if everything is ok but the request to it does not contain a `User-Agent` header so that it looks like the request is coming from a browser.

This made our k8s probes fail for no real reason.

Adding this to the header of the probes solves the problem.

Currently deployed at tasko.dev.renku.ch. I also tested that it does not mess up the probes for regular jupyterlab images.

If you want to test with a rstudio project, there is a public one here: https://tasko.dev.renku.ch/projects/tasko.olevski/test-r-1

This is the discourse post that helped: 
https://community.rstudio.com/t/opening-rstudio-using-aws-ssl-certificate/22006/20

### DIFF
--- a/controller/templates/statefulset.yaml
+++ b/controller/templates/statefulset.yaml
@@ -114,7 +114,7 @@ spec:
               # Required because rstudio sends a 404 even if everything is ok and the user-agent is left out.
               httpHeaders:
                 - name: User-Agent
-                  value: "Mozilla/5.0"
+                  value: "Mozilla/5.0 Firefox"
               path: {{ path }}
               port: 8888
             periodSeconds: 10
@@ -125,7 +125,7 @@ spec:
               httpHeaders:
               # Required because rstudio sends a 404 even if everything is ok and the user-agent is left out.
                 - name: User-Agent
-                  value: "Mozilla/5.0"
+                  value: "Mozilla/5.0 Firefox"
               path: {{ path }}
               port: 8888
             periodSeconds: 10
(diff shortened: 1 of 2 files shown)


controller/templates/statefulset.yaml
```diff
@@ -111,20 +111,32 @@ spec:
           # should be put into an init container.
           livenessProbe:
             httpGet:
+              # Required because rstudio sends a 404 even if everything is ok and the user-agent is left out.
+              httpHeaders:
+                - name: User-Agent
+                  value: "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/71.0.3578.98 Safari/537.36"
```
ableuler, line 117: Do we have to be that specific or do you know if there is some kind of generic "I am a browser" value that Rstudio would accept

olevski: Yeah I just took this from my browser requests. Let me see if there is a more basic version that rstudio will accept.

olevski:
I tested this deployed along with the fix for the probes in Amalthea at https://github.com/SwissDataScienceCenter/renku-notebooks/pull/794.

I tested with an R project that is using these images:

renku/renkulab-r:4.0.5-0.10.1 (i.e. latest, works as expected)
renku/renkulab-r:4.0.5-0.9.0 (In 0.10.1 we added a change to the R image itself that fixes the iframe issue, but now this image version works also because the fix applied here to the session ingress)

ableuler (approved):
Browser spec seems very generic now 👍 .

PR #118, opened 2021-11-11
chore: prep 0.2.0 release

olevski:
So because we have a new feature as well we go to `0.2.0`?

We can also just bump to `0.1.4` instead of going up to `0.2.0`.


ableuler (approved):
0.2.0 is the way to go I'd say

ableuler (approved):

PR #124, opened 2021-11-17
fix(app): ignore the number of kernels when culling sessions

olevski:
The culling currently checks that a session has zero kernels to consider this session idle. I think this is overly generous.

We have some very old sessions that have no activity but 10 kernels.

```
Checking idle status of session anon-XXXXXX, idle seconds: 0, cpu usage: 11.878941m, server status: {'connections': 0, 'kernels': 10, 'last_activity': datetime.datetime(2021, 11, 15, 20, 9, 49, 812356, tzinfo=datetime.timezone.utc), 'started': datetime.datetime(2021, 11, 15, 17, 30, 55, 843291, tzinfo=datetime.timezone.utc)}
```

Therefore, we should not take into account the number of kernels at all when deciding whether a session is idle.

### DIFF
--- a/controller/server_controller.py
+++ b/controller/server_controller.py
@@ -137,7 +137,6 @@ def cull_idle_jupyter_servers(body, name, namespace, logger, **kwargs):
         cpu_usage <= config.CPU_USAGE_MILLICORES_IDLE_THRESHOLD
         and type(js_server_status) is dict
         and js_server_status.get("connections", 0) == 0
-        and js_server_status.get("kernels", 0) == 0
         and (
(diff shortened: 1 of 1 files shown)


ableuler:
I think what we're missing now is at a different level. Rather than culling servers which still have kernels, we should allow the server to cull inactive kernels. See [this post](https://discourse.jupyter.org/t/which-is-the-correct-way-to-cull-idle-kernels-and-notebook/8123/7) for discussion.

ableuler:
Closing this PR in favour of #128 which will handle this more completely.

PR #136, opened 2022-02-02
fix(helm-chart) Fix scheduling through optional scheduler.

aledegano:
* Give a name to the scheduler

 * Fix required RBAC for the scheduler to work properly

### DIFF
--- a/helm-chart/amalthea/templates/scheduler/rbac.yaml
+++ b/helm-chart/amalthea/templates/scheduler/rbac.yaml
@@ -37,9 +37,176 @@ apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: {{ include "amalthea.fullname" . }}-scheduler
 rules:
-  - apiGroups: [""]
-    resources: ["configmaps"]
-    verbs: ["get", "list", "watch"]
+  - apiGroups:
+    - ""
+    resources:
+    - configmaps
+    verbs:
+    - get
+    - list
+    - watch
+  # Copied from https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/main/jupyterhub/templates/scheduling/user-scheduler/rbac.yaml
+  - apiGroups:
+    - ""
+    - events.k8s.io
+    resources:
+    - events
+    verbs:
+    - create
+    - patch
+    - update
+  - apiGroups:
+    - coordination.k8s.io
+    resources:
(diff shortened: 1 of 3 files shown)


helm-chart/amalthea/templates/scheduler/rbac.yaml
```diff
@@ -37,9 +37,176 @@ apiVersion: rbac.authorization.k8s.io/v1
...
+  - apiGroups:
+    - ""
+    resources:
+    - configmaps
+    verbs:
+    - get
+    - list
+    - watch
+  # Copied from https://github.com/jupyterhub/zero-to-jupyterhub-k8s/blob/main/jupyterhub/templates/scheduling/user-scheduler/rbac.yaml
+  - apiGroups:
```
olevski, line 49: @aledegano do we really need all this? I wonder if some of this is JH specific that we do not need.

Reading at the docs on k8s seem to indicate that you can just make a clusterrole binding that adds

```
name: system:kube-scheduler
  apiGroup: rbac.authorization.k8s.io
```

I am wondering if you already tried this? If not I can try it on my deployment since I have this running. Everything else is good.

olevski: This is the section of the docs I am referring to: https://kubernetes.io/docs/tasks/extend-kubernetes/configure-multiple-schedulers/#define-a-kubernetes-deployment-for-the-scheduler

olevski: This is the section of the docs I am referring to: https://kubernetes.io/docs/tasks/extend-kubernetes/configure-multiple-schedulers/#define-a-kubernetes-deployment-for-the-scheduler

aledegano: ah, looking more carefully at my own copy-pasta: the role binding you are suggesting to add it's already there at L15 and unfortunately was not enough as the scheduler would complain of not having enough permissions.

aledegano: I'm reading the documentation a bit better and realized that the way to configure the scheduler has changed and what I've been using here is deprecated.
I'll close this PR while I go back to the drawing board.

olevski:
@aledegano disregard this. I saw your comment.

PR #142, opened 2022-02-16
chore: remove ipython as dev dependency

ableuler:
ipython has a vulnerability and should not really be there as dev dependency in my opinion.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -14,7 +14,6 @@ jinja2 = "*"
 pytz = "*"
 
 [dev-packages]
-ipython = "*"
 black = "*"
 flake8 = "*"
 pytest = "*"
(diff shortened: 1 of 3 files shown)


olevski (changes requested):
Sounds good - I have no clue why or how we even got it in the first place.

Just the linting is failing. After that will approve.

ableuler:
> Sounds good - I have no clue why or how we even got it in the first place.

was probably me putting it there on the initial commit per accident.

ableuler:
hmmm, interesting. Uninstalling ipython downgraded pylint from a pre-release version to the latest released version... Anyway, I've now pinned pylint to the pre-release again, let's hope this fixes things.

olevski (approved):

PR #145, opened 2022-03-10
feat(app): expose metrics to prometheus 

### LINKED ISSUE #143: Expose metrics to be scraped by prometheus
A first set of metrics to publish under a dedicated endpoint could be:

| Metric                          | Type      | Description                                                             |
|---------------------------------|-----------|-------------------------------------------------------------------------|
| sessions_launch_total           | Counter   | Number of sessions launched                                             |
| sessions_all                    | Gauge     | Number of sessions                                                      |
| sessions_running                | Gauge     | Number of sessions in running state                                     |
| sessions_not_scheduled          | Gauge     | Number of which have not been scheduled on any node yet                 |
| sessions_starting               | Gauge     | Number of sessions which have been scheduled but are still starting     |
| sessions_failed_total           | Counter   | Number of sessions which have failed being launched                     |
| session_launch_duration_seconds | Histogram | Time in seconds it takes for a session to transition into running state |
| cpu_request_millicpu            | Histogram | The CPU request of a session in milliCPU                                |
| memory_request_bytes            | Histogram | The memory request of a session in bytes                                |
| disk_request_bytes              | Histogram | The disk request of a session in bytes                                  |

In general we should look out for cardinality. This is explained well in [this article](https://www.robustperception.io/cardinality-is-key). I think defining labels is quite use-case specific. Therefore I think we could could have something like that at the amalthe chart level:

```
metrics:
  enabled: true
  extraMetricsLabels: # A list of labels that must be present on jupyter-server object which will then be used as metrics label
    - renku.io/username
    - renku.io/projectName
```

While for some deployments the above labels might be ok, they won't be for others. Therefore, the default would be to not use any extra metrics labels.

vigsterkr:
Fixes #143

### DIFF
--- a/controller/server_controller.py
+++ b/controller/server_controller.py
@@ -1,7 +1,9 @@
 import kopf
+from enum import Enum
 from kubernetes.client.models import V1DeleteOptions
 from kubernetes.dynamic.exceptions import NotFoundError
 from datetime import datetime
+from prometheus_client import Counter, Gauge
 import pytz
 
 from controller import config
@@ -13,6 +15,60 @@ from controller.utils import (
     get_api,
     parse_pod_metrics,
 )
+
+total_launch = Counter("sessions_launch_total", "Number of sessions launched")
+total_failed = Counter(
+    "sessions_failed_total", "Number of sessions which have failed being launched"
+)
+num_of_sessions = Gauge("sessions_all", "Number of sessions")
+num_of_running_sessions = Gauge(
+    "sessions_running", "Number of sessions in running state"
+)
+num_of_starting_sessions = Gauge(
+    "sessions_starting",
+    "Number of sessions which have been scheduled but are still starting",
+)
+
+
+class SessionState(Enum):
+    """Enum to represent the current state of a session."""
+
+    CREATED = 1
+    STARTING = 2
+    RUNNNING = 3
+    STOPPED = 4
+
+
+sessions_state = dict()
+
+
+def _update_session_state(name, namespace, state):
+    session_key = f"{namespace}_{name}"
+    if session_key not in sessions_state:
+        num_of_sessions.inc()
+        num_of_starting_sessions.inc()
+
+        sessions_state[session_key] = SessionState.CREATED
+        return
+
+    if sessions_state[session_key] == state:
+        return
+
+    if state == SessionState.STARTING:
+        num_of_starting_sessions.dec()
+        sessions_state[session_key] = state
+
+    elif state == SessionState.RUNNNING:
+        num_of_running_sessions.inc()
+        sessions_state[session_key] = state
+
+    elif state == SessionState.STOPPED:
+        num_of_running_sessions.dec()
+        sessions_state[session_key] = state
+
+    else:
+        # we should raise exception here
+        raise ValueError(f"Unknown state type '{state}'!")
 
 
 def get_labels(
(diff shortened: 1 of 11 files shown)


vigsterkr:
> @vigsterkr this is marked as draft. But it seems like it may be ready for review?

@olevski sorry!🤦

controller/server_controller.py
```diff
@@ -14,6 +16,62 @@
...
+    Starting = "starting"
+    Stopping = "stopping"
+    Failed = "failed"
+
+    @classmethod
+    def list(cls):
+        return list(map(lambda c: c.value, cls))
+
+
+sessions_state = dict()
```
olevski, line 46: @vigsterkr you do not need to save a state like this. All the state you need is in the k8s manifests already. And you get this in every handler in kopf automatically.

- for `sessions_launch_total` increment in the existing kopf.on.create below or make another one, never decrement
- for `sessions_all` increment in kopf.on.create, add a new kopf.on.delete handler and decrement there
- for `sessions_starting` increment in kopf.on.create, decrement whenever the state changes from running to something else
- for `sessions_running` increment when status changes from starting to running, decrement when status changes from running to anything else
- for `sessions_failed_total` increment when status goes to failed (from any previous status), never decrement

To compare the old and new status you can simply register a new field handler like this on the jupyterserver resources:
https://kopf.readthedocs.io/en/stable/handlers/#field-handlers. This will give you direct access to the new and old values under `status.state`.

olevski (changes requested):
See my comment below. You can use the existing event handlers or add new event handlers that will trigger exactly when you want to increment/decrement the metrics. There is no need to have any state in the code.

controller/server_controller.py
olevski: Also please note that `update_status` will run for every child of the jupyterServer. (see https://github.com/SwissDataScienceCenter/amalthea/pull/145/files#diff-6346752d3f2c1f6b94bc66266fc9a17eb6efc937693109761d26759d9d0ab5d6R425). But to update the state you need to look at only the `mainPod`. So you do not need to run the checks with every invocation because in this cases the `body` could be anything.

There is already a variable to tell you if you are handling the main pod or not [here](https://github.com/SwissDataScienceCenter/amalthea/pull/145/files#diff-6346752d3f2c1f6b94bc66266fc9a17eb6efc937693109761d26759d9d0ab5d6R425). So just run the state update when `is_main_pod` is true. In all other cases the handler triggered on something that (for now at least) we do not care about. These other children are services, ingresses, secretes, pvc, configmaps.

olevski (dismissed):

aledegano:
This is great!
Should we include here the annotations to the service that allows Prometheus to automatically scrape the metrics (when enabled)?

olevski:
> This is great! Should we include here the annotations to the service that allows Prometheus to automatically scrape the metrics (when enabled)?

@aledegano yeah I can do that.

olevski:
@aledegano prometheus can be setup to scrape pods directly. As a matter of fact that is how ours is setup. The only thing I had to add is allow ingress on the metrics port in the network policy for amalthea. After that everything works:

image


The helm chart for amalthea already allows you to add custom annotations. So that can be used to enable scraping. There is no need to add more stuff in the already crowded values file. And there is no need to create a service.

olevski (approved):

PR #146, opened 2022-03-17
fix: allow number or string for disk size in CRD

olevski:
K8s allows numbers (interpreted as bytes) for memory/disk size.

However in the CRD we ask the user to specify disk size and then pass that on to PVC or emptyDir specs. But the CRD limits this only to strings.

With this we accept both strings and numbers.

I did some tests locally to ensure that PVCs and emptyDirs will accept the values - and they do.


ableuler (approved):
👍

PR #151, opened 2022-03-30
fix(app): use group in dynamic k8s client

olevski:
Some recent issues were possibly related to this.

It was never confirmed whether this would have helped or not. But it is still a good idea to use the API group if available to avoid confusion and problems in the future.

I checked the python k8s client and when the group is not passed in to the `resource.get` function it defaults to `None`

### DIFF
--- a/controller/utils.py
+++ b/controller/utils.py
@@ -189,20 +189,20 @@ def parse_du_command(capacity, bytes_multiplier=1024):
     return None
 
 
-def get_api(api_version, kind):
+def get_api(api_version, kind, group=None):
     """
     Get the proper API for a certain resource. We cache the resources
     availabe in the cluster for 60 seconds in order to reduce the amount
     of unnecessary requests in busy clusters.
     """
     try:
-        return api_cache[(api_version, kind)]
+        return api_cache[(api_version, kind, group)]
     except KeyError:
         client = dynamic.DynamicClient(api_client.ApiClient())
         api_cache[(api_version, kind)] = client.resources.get(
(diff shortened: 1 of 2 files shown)


controller/utils.py
```diff
@@ -189,20 +189,20 @@ def parse_du_command(capacity, bytes_multiplier=1024):
...
     Get the proper API for a certain resource. We cache the resources
     availabe in the cluster for 60 seconds in order to reduce the amount
     of unnecessary requests in busy clusters.
     """
     try:
-        return api_cache[(api_version, kind)]
+        return api_cache[(api_version, kind, group)]
     except KeyError:
         client = dynamic.DynamicClient(api_client.ApiClient())
         api_cache[(api_version, kind)] = client.resources.get(
```
ableuler, line 202: I think you're missing the group in the key here.

controller/utils.py
```diff
@@ -189,20 +189,20 @@ def parse_du_command(capacity, bytes_multiplier=1024):
...
-            api_version=api_version, kind=kind
+            api_version=api_version, kind=kind, group=group,
```
ableuler, line 203: Does it still work for say pods with `group=None`? I'd assume this is the default set in the definition of `client.resources.get` anyway?

ableuler (changes requested):

controller/utils.py
olevski: Yeah. My bad.

ableuler (approved):
Definitely an improvement, even if it should not fix the problem we've seen. 👍

PR #154, opened 2022-04-13
chore: run as non-root

### LINKED ISSUE #153: Run as non-root

olevski:
Run the amalthea operator as non-root.

Adds non-root users to session containers as well as setting `allowPrivilegeEscalations` to false.

closes #153

### DIFF
--- a/controller/templates/statefulset.yaml
+++ b/controller/templates/statefulset.yaml
@@ -149,6 +149,8 @@ spec:
         {% if oidc["enabled"] %}
         - name: oauth2-proxy
           image: "bitnami/oauth2-proxy:7.1.3"
+          securityContext:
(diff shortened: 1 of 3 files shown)


controller/templates/statefulset.yaml
```diff
@@ -102,8 +102,8 @@ spec:
           securityContext:
             runAsUser: 1000
             runAsGroup: 100
-            fsGroup: 100
```
olevski, line 105: `fsGroup` is not allowed in a container security context. Only at the pod level. See [here](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.24/#securitycontext-v1-core).

mohammad-sdsc (approved):
🚀

PR #164, opened 2022-05-25
feat(app): add and modify prometheus metrics

### LINKED ISSUE #156: Finish implementing all Amalthea prometheus metrics
There are a few metrics left over to be implemented from https://github.com/SwissDataScienceCenter/amalthea/issues/143.

Implement all of them.

olevski:
Modifies the existing metrics from #143 and adds the rest of the metrics that were note initially implemented.

To avoid counting and also therefore potentially double-counting the number of active sessions all such metrics were removed. The number of active sessions can be estimated by looking at the differences between session launches and shutdowns.

If the number of active sessions is tracked then we need some way of tracking which sessions have been counted or not to avoid double counting. Also we need to be able to track when initializing the metric whether the sessions that are initially there have been counted or not.

With https://github.com/SwissDataScienceCenter/renku-notebooks/issues/1069 the proper and long term storage of detailed metrics will be designed and resolved.

closes #156

### DIFF
--- a/controller/culling.py
+++ b/controller/culling.py
@@ -1,9 +1,10 @@
+from dateutil import parser
 from json.decoder import JSONDecodeError
 import logging
 import requests
 from requests.exceptions import RequestException
 
-from controller.utils import get_pod_metrics, parse_pod_metrics, k8s_timestamp_to_utc_datetime
+from controller.utils import get_pod_metrics, parse_pod_metrics
 
 
 def get_cpu_usage_for_culling(pod, namespace):
@@ -73,7 +74,7 @@ def get_js_server_status(js_body):
         return None
 
     if type(res) is dict and "last_activity" in res.keys():
-        res["last_activity"] = k8s_timestamp_to_utc_datetime(res["last_activity"])
+        res["last_activity"] = parser.isoparse(res["last_activity"])
     if type(res) is dict and "started" in res.keys():
-        res["started"] = k8s_timestamp_to_utc_datetime(res["started"])
+        res["started"] = parser.isoparse(res["started"])
     return res
--- a/controller/culling.py
+++ b/controller/culling.py
@@ -1,11 +1,9 @@
-from datetime import datetime
 from json.decoder import JSONDecodeError
 import logging
-import pytz
 import requests
 from requests.exceptions import RequestException
 
-from controller.utils import get_pod_metrics, parse_pod_metrics
+from controller.utils import get_pod_metrics, parse_pod_metrics, k8s_timestamp_to_utc_datetime
 
 
 def get_cpu_usage_for_culling(pod, namespace):
@@ -75,19 +73,7 @@ def get_js_server_status(js_body):
         return None
 
     if type(res) is dict and "last_activity" in res.keys():
-        res["last_activity"] = datetime.fromisoformat(
-            res["last_activity"][:-1] + "+00:00"
-            if res["last_activity"].endswith("Z")
-            else res["last_activity"]
-        ).astimezone(
-            pytz.utc
-        )  # ensure timestamp is UTC
+        res["last_activity"] = k8s_timestamp_to_utc_datetime(res["last_activity"])
     if type(res) is dict and "started" in res.keys():
-        res["started"] = datetime.fromisoformat(
-            res["started"][:-1] + "+00:00"
-            if res["started"].endswith("Z")
-            else res["started"]
-        ).astimezone(
-            pytz.utc
-        )  # ensure timestamp is UTC
+        res["started"] = k8s_timestamp_to_utc_datetime(res["started"])
     return res
--- a/controller/metrics.py
+++ b/controller/metrics.py
@@ -1,4 +1,5 @@
 from dataclasses import dataclass
+import logging
 from prometheus_client import Counter, Histogram
 from typing import Dict, Union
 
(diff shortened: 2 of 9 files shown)


controller/culling.py
```diff
@@ -75,19 +73,7 @@ def get_js_server_status(js_body):
         return None
 
     if type(res) is dict and "last_activity" in res.keys():
-        res["last_activity"] = datetime.fromisoformat(
```
olevski, line 78: moved this code into its own function because I need to use it in a few cases.

controller/utils.py
```diff
@@ -257,3 +261,13 @@ def sanitize_prometheus_metric_label_name(val):
     val = re.sub(first_letter, "_", val, count=1)
     val = re.sub(all_letters, "_", val)
     return val
+
+
+def k8s_timestamp_to_utc_datetime(timestamp: str) -> datetime:
+    return datetime.fromisoformat(
```
Panaetius, line 267: Did you consider using https://dateutil.readthedocs.io/en/stable/parser.html#dateutil.parser.isoparse ? It's what https://docs.python.org/3/library/datetime.html#datetime.datetime.fromisoformat recommends for parsing arbitrary iso8601 strings. 

Not sure what the risk of getting a format that this code here doesn't cover is.

controller/server_controller.py
```diff
@@ -128,23 +130,35 @@ def delete_fn(labels, body, namespace, name, **_):
...
-    current_state = body.get("status", {}).get("state", ServerStatusEnum.Starting.value)
     api = get_api(config.api_version, config.custom_resource_name, config.api_group)
+    new_status = ServerStatusEnum.Stopping.value
+    if body:
+        old_status = body.get("status", {}).get("state")
+    else:
+        old_status = None
+    if old_status != new_status:
+        extra_labels = {
+            "status_from": str(old_status),
```
Panaetius, line 141: Should we lower-case this? All the `ServerStatusEnum` values are lower-case. I also wonder if we want to have a dedicated `indeterminate` or `unknown` value instead of `none` that we could use here.


There's also the philosophical question on if going from no status to a status constitutes a status change (i.e. if we should count it at all), but then I think it doesn't make a difference. :slightly_smiling_face:

controller/server_controller.py
```diff
@@ -153,22 +167,50 @@ def delete_fn(labels, body, namespace, name, **_):
...
+        now = pytz.UTC.localize(datetime.utcnow())
+        start_duration = (now - start_time).total_seconds()
+        metrics.manipulate(
+            "sessions_launch_duration",
+            "observe",
+            start_duration,
+            labels,
+        )
+
+    if new == ServerStatusEnum.Starting.value:
```
Panaetius, line 187:
```suggestion
    elif new == ServerStatusEnum.Starting.value:
```

controller/utils.py
olevski: > Did you consider using https://dateutil.readthedocs.io/en/stable/parser.html#dateutil.parser.isoparse

I had no clue about this... I will use this instead. I think that the timestamps in k8s will either have an offset in hours or just `Z` for UTC. So The risk of getting a code that is not covered is low - or def lower than this hacky function.

Panaetius (approved):

Panaetius (approved):

olevski:
@Panaetius can you re-approve this please.

I found a few issues:
- Updating a prometheus metric was failing causing the operator to fail also to fully delete everything. I fixed this and made sure that if using a metric fails that the operator does not error out.
- The PR we recently merged that introduced culling for idling and starting sessions has some issues it seems - namely that startingSince and failingSince data is never patched into the jupyterserver status. So the sessions never get culled, I will open a bug for this and fix it in another PR. This was not part of this PR anyhow.

olevski:
https://github.com/SwissDataScienceCenter/amalthea/pull/164

controller/metrics.py
```diff
@@ -1,56 +1,88 @@
...
-            labels = {}
             for i in range(len(config.METRICS_EXTRA_LABELS)):
                 try:
                     labels[config.METRICS_EXTRA_LABELS_SANITIZED[i]] = manifest_labels[
                         config.METRICS_EXTRA_LABELS[i]
                     ]
                 except KeyError:
                     # NOTE: We should not be super sensitive to k8s label changes, so if a
-                    # lable that is exepected to exist in the manifest does not then
+                    # label that is exepected to exist in the manifest does not then
```
Panaetius, line 73: `exepected`

controller/metrics.py
```diff
@@ -1,56 +1,88 @@
...
                     # the metric label will not be applied, but the metric will be counted
                     labels[config.METRICS_EXTRA_LABELS_SANITIZED[i]] = "Unknown"
+        labels = {**labels, **extra_labels}
+        if len(labels.keys()) > 0:
             metric = metric.labels(**labels)
-        getattr(metric, operation)()
+        try:
+            getattr(metric, operation)(value)
```
Panaetius, line 80: Should we first check `getattr(metric, operation, None) is not None`, with an error log entry like "Operation xyz doesn't exist" and only then do this try&except ?

Panaetius, line 80: Should we first check `getattr(metric, operation, None) is not None`, with an error log entry like "Operation xyz doesn't exist" and only then do this try&except ?

Panaetius (changes requested):

controller/metrics.py
olevski: 🤦

olevski: thanks

Panaetius (approved):

PR #171, opened 2022-06-16
chore: add custom CA certificate injection

olevski:
To properly do idle session culling amalthea needs to send a get request to `https://<ful_session_path>/api/status`.

In cases where self signed certificates are used this would fail without injecting these in amalthea.


olevski:
`<full_session_path>` refers to what is usually the combination of `<hostname>/sessions/<sesion_id>` where the session can be accessed. I.e. it is the combination of the hostname and path in the ingress object created by amalthea.

Panaetius (approved):
Looks good

PR #173, opened 2022-06-23
feat(app): long term metric storage

olevski:
Adds long term metric storage in a s3 bucket.

Handles all metrics in a separate thread.

closes https://github.com/SwissDataScienceCenter/renku-notebooks/issues/1069


controller/metrics/s3.py
```diff
@@ -0,0 +1,145 @@
...
+        self._current_log_start_timestamp = None
+        self._current_log_id = uuid4()
+        self.__file_object = None
+
+    @property
+    def _file_object(self):
+        if not self._current_log_start_timestamp:
+            self._current_log_start_timestamp = pytz.UTC.localize(datetime.utcnow())
+        now = pytz.UTC.localize(datetime.utcnow())
+        if (now - self._current_log_start_timestamp).total_seconds() / 3600 > self.period_hours:
```
Panaetius, line 108:
```suggestion
        if (now - self._current_log_start_timestamp)/timedelta(hours=1) > self.period_hours:
```
is what the datetime docs recommend for units other than seconds

controller/metrics/s3.py
```diff
@@ -0,0 +1,145 @@
...
+                else None
+            ),
+            metric_event.old_status,
+            manifest.get("metadata", {}).get("annotations", {}).get("renku.io/commit-sha"),
+            manifest.get("metadata", {}).get("annotations", {}).get("renku.io/repository"),
+            manifest.get("metadata", {}).get("annotations", {}).get("renku.io/username"),
+        )
+
+
+class BaseMetricLogger(ABC):
```
Panaetius, line 77: Did you consider creating a new python `logging` handler for S3 instead? The Info/Debug/Warn model isn't that appropriate for this use case, but it would mean being compatible with many other logging implementations, all the features already available like filtering and I think the basic design of pythons logging is done quite well. At least the shouldRollover/doRollover logic in https://github.com/python/cpython/blob/11e865c74baa45ac9eda69f2f6af05505fc03ce5/Lib/logging/handlers.py#L65 is pretty readable.

I think all that'd be needed is to implement `emit` in a class inheriting from https://github.com/python/cpython/blob/11e865c74baa45ac9eda69f2f6af05505fc03ce5/Lib/logging/__init__.py#L919 and to add the shouldRollover/doRollover logic mentioned above.

But I can see arguments for and against, so just wanted to check if you looked into this. Not something you have to add.

controller/metrics/s3.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+class S3MetricHandler(MetricEventHandler):
+    def __init__(self, logger: BaseMetricLogger):
+        self.logger = logger
+
+    def publish(self, metric_event: MetricEvent):
+        session_metric_data = SesionMetricData.from_metric_event(metric_event)
+        self.logger.log(session_metric_data.to_json())
+        self.logger.log("\n")
```
Panaetius, line 145:
```suggestion
        self.logger.log(f"{session_metric_data.to_json()}\n")
```
or
```suggestion
        self.logger.log(session_metric_data.to_json() + "\n)
```
no reason to write twice

Panaetius (changes requested):
We are really lacking class/method docstrings in amalthea, would be nice to have some descriptions on what things do to make them more easily understandable.

controller/metrics/prometheus.py
```diff
@@ -0,0 +1,227 @@
...
+        },
+        "histogram": {
+            "usage_methods": ["observe"],
+            "type": Histogram,
+        },
+        "summary": {
+            "usage_methods": ["observe"],
+            "type": Summary,
+        },
+    }
```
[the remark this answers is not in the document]
olevski: Yeah enum is def better. Added it.

controller/metrics/prometheus.py
```diff
@@ -0,0 +1,227 @@
...
+        val = re.sub(invalid_first_letter, "_", val, count=1)
+        val = re.sub(invalid_all_letters, "_", val)
+        return val
+
+    @classmethod
+    def _sanitize_labels(cls, labels: Dict[str, str]) -> Dict[str, str]:
+        sanitized_labels = {}
+        for name, val in labels.items():
+            sanitized_labels[cls.sanitize_label_name(name)] = val
+        return sanitized_labels
```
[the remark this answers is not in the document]
olevski: I added that because they do did not really need `self` of any of the real state of the object. But with some of the other proposed changes now I do need access to the state and there is no more classmethod or staticmethod

controller/metrics/prometheus.py
```diff
@@ -0,0 +1,227 @@
...
+            )
+        operation_method(value)
+
+
+class PrometheusMetricHandler(MetricEventHandler):
+    def __init__(self, manifest_labelnames: List[str] = []):
+        self.manifest_labelnames = manifest_labelnames
+        self._sessions_total_created = PrometheusMetric(
+            "counter",
+            "sessions_total_created",
```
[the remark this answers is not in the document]
olevski: I can convert the names of the metric to an enum. But I think there is no need tbh. They are all defined as class variables at the top and are accessed only through the class variables. 

Should I just make this an enum of metric names and make that a class variable?

controller/metrics/s3.py
olevski: Now I just use seconds. It is easier that way and it allows me to setup a very short period for testing and development.

controller/metrics/s3.py
olevski: Yeah I found the docs on `logging` really hard to navigate. But I am pretty sure I figured things out. I have a logging handler that uploads to s3 on rotation. And I have a formatter as well. But yeah as you suggest using logging for this is much nicer.

controller/metrics/s3.py
olevski: This is not an issue now with the logging handler and formatter.

controller/metrics/prometheus.py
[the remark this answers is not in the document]
olevski: Ok I just made a simple Enum that contains the different metric names. I see what you mean this way there is pretty much zero chance that we mistype a metric name and inadvertently create a new metric. If we were to mistype a metric name then the Enum would error out.

controller/metrics/queue.py
```diff
@@ -0,0 +1,29 @@
...
+    def __init__(self, metric_handlers=List[MetricEventHandler]):
+        self.q = Queue()
+        self.metric_handlers = metric_handlers
+        self.thread = None
+
+    def _queue_worker(self):
+        while True:
+            metric_event = self.q.get()
+            for handler in self.metric_handlers:
+                handler.publish(metric_event)
```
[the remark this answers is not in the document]
olevski: I added a blanket `except Exception` statement to catch any exception log it as a warning and continue.

olevski:
@Panaetius I addressed all the comments. I also added docstrings to all classes and their public methods in the new files for metrics handling.

I just have to fix the tests and that is it.

controller/metrics/prometheus.py
```diff
@@ -157,67 +197,62 @@ def _collect_labels_from_manifest(self, manifest: Dict[str, Any]) -> Dict[str, s
 
     def _on_start(self, metric_event: MetricEvent):
         manifest_labels = self._collect_labels_from_manifest(metric_event.session)
-        self._sessions_total_created.manipulate("inc", 1, manifest_labels)
+        self._sessions_total_created("inc", 1, manifest_labels)
```
Panaetius, line 200:
```suggestion
        self._sessions_total_created(PrometheusMetricAction.inc, 1, manifest_labels)
```

controller/metrics/prometheus.py
```diff
@@ -157,67 +197,62 @@ def _collect_labels_from_manifest(self, manifest: Dict[str, Any]) -> Dict[str, s
...
+                "status_to": metric_event.status.value if metric_event.status else "None",
             }
-            self._sessions_status_changes.manipulate("inc", 1, status_change_labels)
+            self._sessions_status_changes(PrometheusMetricAction.inc, 1, status_change_labels)
 
     def publish(self, metric_event: MetricEvent):
+        """Publishes (i.e. persists) the proper prometheus metrics
+        depending on the old and new statuses of the jupyterserver."""
         old_status = metric_event.old_status
         new_status = metric_event.session.get("status", {}).get("state")
```
Panaetius, line 255:
```suggestion
        new_status = metric_event.status
```
?

Panaetius (changes requested):
Thank you, it looks really nice now. Glad that using `logging` turned out to be quite easy :+1: 

Two minor places where I think you forgot to change to the new way of doing things.

controller/metrics/prometheus.py
Panaetius: Not sure about this one, `metric_event.session.get("status", {}).get("state")` might be something else. Ignore if it is

controller/config_types.py
```diff
@@ -0,0 +1,66 @@
...
+    endpoint: str
+    bucket: str
+    path_prefix: str
+    access_key_id: str
+    secret_access_key: str
+    rotation_period_seconds: Optional[str] = None
+
+    def __post_init__(self):
+        if not self.rotation_period_seconds:
+            self.rotation_period_seconds = 86400
```
Panaetius, line 19: You can set the default value above instead with `    rotation_period_seconds: Optional[str] = 86400`

controller/config_types.py
```diff
@@ -0,0 +1,66 @@
...
+    enabled: Union[str, bool] = False
+    extra_labels: Union[str, List[str]] = None
+
+    def __post_init__(self):
+        if type(self.enabled) is str:
+            self.enabled = self.enabled.lower() == "true"
+        if type(self.extra_labels) is str:
+            self.extra_labels = json.loads(self.extra_labels)
+        if not self.extra_labels:
+            self.extra_labels = []
```
Panaetius, line 36: Same here

controller/metrics/s3.py
```diff
@@ -41,9 +32,11 @@ class SesionMetricData:
...
     status: Optional[ServerStatusEnum]
     old_status: Optional[ServerStatusEnum]
-    commit: Optional[str]
-    repository_url: Optional[str]
-    user: Optional[str]
+    additional_labels: Optional[Dict[str, str]]
+
+    def __post_init__(self):
+        if not self.additional_labels:
+            self.additional_labels = {}
```
Panaetius, line 39: same here I think

controller/metrics/utils.py
```diff
@@ -20,13 +20,15 @@ def resource_request_from_manifest(manifest: Dict[str, Any]) -> Optional[Resourc
     resources = manifest.get("spec", {}).get("jupyterServer", {}).get("resources", {}).get(
         "requests", {}
     )
+    resources = {**resources}
```
Panaetius, line 23: This is essentially a nop? or what does this do?

If it is just to cash to a plain python dict, `resources = dict(resources)` communicates the intent better, imo.

Panaetius (changes requested):

controller/metrics/utils.py
olevski: @Panaetius yeah I essentially wanted a copy of that dictionary. Because in a few lines below I add something to it just for the purposes of this function. But if I do not use a copy then that same dictionary in the manifest gets modified in the same way and breaks things.

Now I just use `copy.deepcopy` for this which is the safest option and properly communicates the intent IMO.

olevski:
@Panaetius this is ready now. I went with `dict(**resources)` on making a dictionary copy because `deepcopy` failing to create a copy of resources because in some cases it is a plain dictionary (this works) when the session is first created. But after the session is created and the status changes then this is a dict-like object that I get from querying the k8s api. With `dict` both cases work.

Panaetius (approved):

PR #180, opened 2022-07-07
chore: allow showing hidden files

lorenzo-cavazzi:
Allow showing hidden files in the JupyterLab file browser.

### DIFF
--- a/controller/templates/configmap.yaml
+++ b/controller/templates/configmap.yaml
@@ -26,6 +26,7 @@ data:
     c.ServerApp.root_dir="{{ jupyter_server["rootDir"] }}"
     c.ServerApp.default_url="{{ jupyter_server["defaultUrl"] }}"
     c.ServerApp.allow_remote_access=True
+    c.ContentsManager.allow_hidden=True
 
   {% if oidc["enabled"] %}
   authorized-users.txt: |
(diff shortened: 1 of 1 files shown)


lorenzo-cavazzi:
This seems to work, but I don't see the toggle to show/hide hidden files as mentioned here https://jupyterlab.readthedocs.io/en/stable/user/files.html#displaying-hidden-files
The hidden files are always visible.

Screenshot_20220707_171343

Screenshot_20220707_171513

olevski (dismissed):

lorenzo-cavazzi:
Using the latest version of our RenkuLab Docker images (i.e. starting from a newer version of the Juypter Docker images with JupyterLab 3.3.4 ) solves the previous problem: the files are hidden by default and the user can tick a new "Show Hidden Files" option in the "View" menu  :tada: 

It's testable on https://lorenzo.dev.renku.ch using the following custom template:
* Repository: `https://github.com/SwissDataScienceCenter/renku-project-template`
* Branch: `000-bump-images-version`

I'll tag a new version of the templates from https://github.com/SwissDataScienceCenter/renku-project-template/pull/146

Screenshot_20220808_114252

olevski (approved):

olevski:
@lorenzo-cavazzi there is no `#deploy persist` here. The acceptance tests run jupyter servers and test simple scenarios. If those pass then things are fine.

If you want to test with renku then you can do that by deploying yourself or publishing a helm chart and doing a PR in renku-notebooks.

olevski (approved):

lorenzo-cavazzi:
We updated the template images a while ago, I think we can now consider merging this (unless there are reasons to hold-on @olevski )

PR #189, opened 2022-08-08
fix: buckets for prometheus histogram metrics

### LINKED ISSUE #186: Histogram type metrics need sensible buckets
The buckets that get created for the histogram type prometheus metrics do not make sense relative to the values observed.

We should set sensible buckets for these.

I.e. the buckets for memory can have splits at 1GB, 4, 8, 16, 64 and 64+ or something like that.

Similarly we measure cpu requests in millicores and we can have sensible breakdown for 1000, 2000, 4000, 8000, 8000+ millicores

This should be also done for storage and for startup time. Checking the average startup time can help. It is probably somehwere around 2-3 minutes.

olevski:
closes #186 

The default bucket values are so low that they are useless. I had hopes that maybe prometheus automatically assigns buckets based on the values it sees but it make sense that this is not the case.

### DIFF
--- a/controller/metrics/prometheus.py
+++ b/controller/metrics/prometheus.py
@@ -158,6 +158,7 @@ class PrometheusMetricHandler(MetricEventHandler):
             "How long did it take for a session to transition into running state",
             self.manifest_labelnames,
             unit="seconds",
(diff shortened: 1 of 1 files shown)


aledegano (approved):
LGTM!

PR #190, opened 2022-08-08
fix: do not publish same metric more than once

### LINKED ISSUE #185: Metrics get published multiple times when a session is failing
Amalthea will keep retrying sessions that have invalid k8s manifests.

This means that the metrics that are published from certain handlers are published over and over if there is an error.

One way to handle this is by using some annotations that prevent certain handlers from kopf from running again.

https://kopf.readthedocs.io/en/stable/tips-and-tricks/#excluding-handlers-forever

A useful thing to do here is to also move all metrics into their own separate and dedicated handler. Ideally just one handler but if not more than one. Then this handler will only ever run once if there is an error and not be retried multiple times.

olevski:
The `on.create` handler gets repeated indefinitely if the handler function fails. This was causing some metrics from the handler to be published multiple times.

Now with this change the handlers for metrics are separated from the rest. And in addition the metrics are published only when the status of the server changes. So there is no chance that the same metric (for the exact same status change) will be published more than once.

closes #185

### DIFF
--- a/controller/server_controller.py
+++ b/controller/server_controller.py
@@ -617,8 +617,8 @@ if config.JUPYTER_SERVER_RESOURCE_CHECK_ENABLED:
     )(update_resource_usage)
 
 
-# INFO: Start the prometheus metrics server if enabled
-if config.METRICS.enabled:
+# INFO: Register the functions that publish metric events into the metric events queue
+if config.METRICS.enabled or config.AUDITLOG.enabled:
     kopf.on.field(
         config.api_group,
         config.api_version,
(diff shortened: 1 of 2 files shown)


controller/server_controller.py
```diff
@@ -625,6 +590,24 @@ def update_resource_usage(body, name, namespace, **kwargs):
         pass
 
 
+def publish_metrics(old, new, body, name, **_):
+    """Handler to publish prometheus and auditlog metrics on server status change."""
+    if old == new:
+        # INFO: This is highly unlikely to occur, but if for some reason the status has changed
```
Panaetius, line 596:
```suggestion
        # INFO: This is highly unlikely to occur, but if for some reason the status hasn't changed
```

Panaetius (changes requested):
Looks great, just a typo in a docstring :slightly_smiling_face:

Panaetius (approved):

PR #243, opened 2022-09-12
fix(app): do not use previous status for state

olevski:
I think when I added this (originally to the notebook service) I misunderstood what "lastState" meant. 

What it truly means is that when a container is for example killed for consuming too much memory, then the last state will show "OOMKilled" as the reason and it will show the last exit code.

So with the old logic when a container was "OOMKilled" (i.e. restarted for consuming memory above its limit) then it would have it's last state as that. And the logic for the container state would be stuck on "failing" even though the new current state is all good and the whole sessions is functioning normally.

fixes https://github.com/SwissDataScienceCenter/renku-notebooks/issues/1225


olevski:
I confirmed that I can replicate the behaviour in the bug. And that after removing this logic the state reacts as expected and switches back to "running" after the container successfully restarts.

Panaetius (approved):

PR #255, opened 2022-09-22
fix: check url when setting status

olevski:
It can happen that even when all k8s statuses and states are good and the session is marked as "Running" that the service/ingress are still not fully synced. And accessing the server at this time results in a 503 just as if you had accessed the server before the readiness probes all cleared.

I tried many different ways to avoid this but I would always (one way or another) end up with a 503 screen even though the state is running. 

Therefore now I do a GET request to the session URL when I think the server is ready. And if this GET request suceeds the server is marked as ready. If not the request is retried several times before the sessions is marked as "failed".


olevski:
superseded by https://github.com/SwissDataScienceCenter/amalthea/pull/258

PR #258, opened 2022-09-26
feat: detailed status messages

olevski:
Provide by-container summary of the status.

### DIFF
--- a/controller/server_status.py
+++ b/controller/server_status.py
@@ -1,0 +1,316 @@
+"""Classes for working with pod and container states.
+
+Usually deals with data like the following:
+    containerID: docker://XXXX
+    image: some/image:1.2.3
+    imageID: docker-pullable://renku/renku-gateway@sha256:XXXX
+    lastState:
+      terminated:
+        containerID: docker://XXXX
+        exitCode: 0
+        finishedAt: "2022-09-07T11:12:19Z"
+        reason: Completed
+        startedAt: "2022-09-07T11:11:10Z"
+    name: gateway
+    ready: true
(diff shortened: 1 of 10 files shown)


controller/server_status.py
```diff
@@ -0,0 +1,316 @@
...
+        for status in self.statuses:
+            if status.failed:
+                failed_statuses.append(status)
+            elif status.running_ready or status.completed_successfully:
+                ok_statuses.append(status)
+        for status in self.init_statuses:
+            if status.failed:
+                failed_init_statuses.append(status)
+            elif status.running_ready or status.completed_successfully:
+                ok_init_statuses.append(status)
```
Panaetius, line 301: The list here is not actually used, it could just be ints that get incremented. Also, could `self.statuses` and `self.init_statuses` be chained/combined and just a single for loop used? And then below it'd just be something like

`and num_ok_statuses == len(self.statuses)+len(self.init_statuses)` ? Actually you could just count total number of statuses in the for loop and not use `len` at all.

Panaetius (changes requested):
Thank you, looks good. The logic for the different statuses is a bit hard to follow in places, but then I'm not sure how it could be simplified/more self-explanatory.

There's a typo in a comment and once suggestion that you can ignore.

controller/server_status.py
olevski: @Panaetius I think I originally had some other plans with those and that is why I had the arrays. Anyhow you are right all that stuff is not needed.

Panaetius (approved):

PR #283, opened 2022-10-28
feat(chart): refactor additional namespaces setup

olevski:
This changes the meaning of the "namespaces" value in the values files.

The original meaning is that if amalthea can be deployed either in "cluster" mode or namespaced. In cluster mode amalthea gets a clusterrole and it can create servers anywhere. In namespaced mode we allow several namespaces to be defined and we create the required resources in each namespace. The values file is setup such that if the `.Values.scope.namespaces` field is undefined then amalthea has permissions to create servers only in the namespace where it is deployed. But if this value is specified then amalthea has permissions in the namespace where the helm chart is deployed and any namespaces from the list in the values file. This is slightly cumbersome because we have to keep the field undefined to get the default and then define it to get additional namespaces.

Now I changed this field so that it is called `additionalNamespaces`. This is now defined and its default value in the values file is an empty list. And if left like that then amalthea has permissions only for the namespace where the helm chart is deployed. But if values are added to the `additionalNamespaces` field then amalthea has access to these namespaces in addtion to the namespace where the helm chart is deployed.

I think this makes things easier because we can now specify additional namespaces without bothering to include the helm chart namespace in the list. Before this if we wanted to give amalthea permissions to the namespace where the helm chart is deployed and 2 more we would have to list the namespace where the helm chart is deployed in this list in the values file. This is redundant and weird because this is not usually done. The helm chart knows and can access the value of the namespace where it will be deployed so it should not be specified in the values file.

### DIFF
--- a/helm-chart/amalthea/templates/deployment.yaml
+++ b/helm-chart/amalthea/templates/deployment.yaml
@@ -50,7 +50,7 @@ spec:
             {{- if .Values.scope.clusterWide }}
             - "--all-namespaces"
             {{- else }}
-            {{- range (.Values.scope.namespaces | default (list .Release.Namespace)) }}
+            {{- range (append .Values.scope.additionalNamespaces .Release.Namespace | uniq) }}
             - "--namespace={{ . }}"
             {{- end }}
             {{- end }}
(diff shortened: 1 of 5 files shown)


Panaetius:
Is there a use-case where some deployment would ONLY want session to be available in a separate namespace? Like maybe nodes with GPUs for sessions only. So where they do not want sessions to be run in the default namespace?

With this implementation, that would have to be restricted in the UI or notebook service, right?

olevski:
@Panaetius I will close this. I thought it was going to be needed for more easily supporting multiple namespaces in the notebook service. But keeping the helm chart as is is fine.

And we also make it possible for someone to make sessions available only in a separate namespace.

PR #336, opened 2023-06-16
feat: add error when resource quota is exceeded

olevski:
With the addition of resource quotas this is the only way to surface a message that a k8s resource quota is exceeded when we try to schedule a jupyter server.


mohammad-alisafaee (approved):
LGTM, thank you!

PR #344, opened 2023-07-21
feat(app): use pv to persist sessions

mohammad-alisafaee:
Adds a KOPF field watcher for `spec.jupyterServer.hibernated` that scales a server's statefulset replicas to 0/1 depending if the server is being hibernated/resumed. Server's `state` is set to `hibernated` if `spec.jupyterServer.hibernated` is true.

# NOTE:
- `JupyterServer` CRD in this PR needs to be applied to the `dev` cluster for this to work. You can either set `notebooks.amalthea.deployCrd` set to true in your values file or apply `manifests/crd.yaml` in this PR to the `dev` cluster. Note that CRDs aren't namespaced in K8s and this will affect the whole cluster and the `dev` deployment. This seems to be causing https://swiss-data-science.slack.com/archives/G67N59QL8/p1689682274613449.
- You need to set `notebooks.userSessionPersistentVolumes.enabled` to true in your values files to enable the use of PV for Jupyter servers.

Related: https://github.com/SwissDataScienceCenter/renku-notebooks/pull/1518.

### DIFF
--- a/.github/workflows/acceptance-tests.yaml
+++ b/.github/workflows/acceptance-tests.yaml
@@ -109,6 +109,7 @@ jobs:
           TEST_IMAGE_NAME: ${{ matrix.test-image.image }}
           TEST_SPEC: ${{ matrix.test-image.spec }}
           ENVIRONMENT: ${{ matrix.test-image.env }}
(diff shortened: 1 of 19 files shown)


olevski (approved):
Thanks Mohammad!

PR #457, opened 2023-11-15
move chartpress yaml to work with deploy action

Panaetius:


chartpress.yaml
```diff
@@ -0,0 +1,14 @@
+charts:
```
olevski, line 1: There is already a chartpress file in helm-chart/amalthea

controller/config.py
```diff
@@ -27,6 +27,7 @@
     {"name": "configmaps", "group": ""},
     {"name": "services", "group": ""},
     {"name": "persistentvolumeclaims", "group": ""},
+    {"name": "persistentvolumes", "group": ""},
```
olevski, line 30: I dont think we should merge this in and release. Lets wait until we have our own version of the controller that does not need access to PVs. I thought that was the plan?

helm-chart/amalthea/values.yaml
```diff
@@ -33,6 +33,10 @@ scope:
   # 2. Do not define namespaces at all, in which case amalthea
   #    will only operate in the namespace where the helm chart is deployed.
   # namespaces: ["default"]
+  clusterRoleName: amalthea
+  # name of the cluster role for PV creation
+  # only needs to be set if not deployed clusterwide
+  # the role needs permission to create,get,list,watch PVs
```
olevski, line 39: The info/docs should be above the field they are explaining. Namespaces above is a field and is commented out. So we should move this above the field to avoid confusion. But hopefully we won't need to use this at all on account of the new CSI driver.

olevski (changes requested):
I would rather keep this in a feature branch until we have the new CSI driver ready that does not require access to PVs.

controller/config.py
Panaetius: :+1: 
yes that's why it's still a draft

chartpress.yaml
olevski: I dont know what the deploy action needs. 

Either way we should definitely not make copies of this file. Either move the file to where the actions are happy or symlink.

olevski (approved):

PR #494, opened 2024-01-16
fix: status can only be hibernated when server is hibernated

olevski:
This is a small fix for a rare bug that Flora noticed.

Basically what could happen is that when you hibernate a session that is failed sometime the function that was modified below would flip the status back to "Failed" but the status should stay "Hibernated".

This fixes the problem.

### DIFF
--- a/controller/server_status.py
+++ b/controller/server_status.py
@@ -292,6 +292,8 @@ class ServerStatus:
     @property
     def overall_status(self) -> ServerStatusEnum:
         """Get the status of the jupyterserver."""
+        if self.hibernated:
+            return ServerStatusEnum.Hibernated
         if self.deletion_timestamp:
             return ServerStatusEnum.Stopping
         if self.is_unschedulable:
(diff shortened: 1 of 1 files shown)


controller/server_status.py
```diff
@@ -292,6 +292,8 @@ def server_url_is_eventually_responsive(self, timeout_seconds: int = 5) -> bool:
     @property
     def overall_status(self) -> ServerStatusEnum:
         """Get the status of the jupyterserver."""
+        if self.hibernated:
```
mana-alisafaee, line 295: This exists in the following lines (line 306). Was it a reason to move it here? I believe the other two checks have higher priority than this one. Anyways we should remove the other one if this is required here.

mana-alisafaee (changes requested):

Panaetius (approved):

controller/server_status.py
olevski: Ok I did not see that was already there. We should move it up because what Flora saw was that a session was failing because it was unschedulable. Then it was hibernated (in this unschedulable state) and the status went from failing to hibernated (as expected) but then after a few seconds it flipped back to failed/unschedulable. By moving this check up we avoid the problem. If the session is hibernated we should get the hibernated status regardless of what was going on when it was running. Especially now that we can modify the session when it is hibernated the old failing status is really not important.

mana-alisafaee (approved):

PR #580, opened 2024-06-20
feat(networkpolicies): customize cluster IPs

JGodin-C2C:
# What

Allows a user to customize his network according to internal needs

# Why 
 
In case a user have another network with private Ips that needs to be reached, this allow the user to deploy a more custom networkpolicy for more flexibility


aledegano (approved):
LGTM

PR #582, opened 2024-06-23
chore: Scaffold go operator framework

olevski:
Note that the target branch is a feature branch.


ableuler:
Wow - somebody has put that rainy weekend to very good use! 🚀 
Happy to see that this project is still alive and kicking, even though this PR suggests that you're going to refactor my first shot things out of the code base entirely 😅. Excited to see where you are taking this project!

olevski:
> even though this PR suggests that you're going to refactor my first shot at things out of the code base entirely 😅

Hi @ableuler! Your version of the operator is not going anywhere (for now). We just want to add another CRD and another operator that is more general and not tied to Jupyter. Then eventually we can retire your initial version or  convert it to Go.

> somebody has put that rainy weekend to very good use

And yeah the weather on the weekend was right for staying indoors and writing k8s operators. Haha.

olevski:
Ah also @ableuler 99% of that code is auto-generated by the `operator-sdk` CLI.

olevski:
@sgaist since you played with this already it would be great to merge this if you approve. It will not affect the existing kopf  operator at all.

sgaist:
> @sgaist since you played with this already it would be great to merge this if you approve. It will not affect the existing kopf operator at all.

From a code point of view I agree however the `Dockerfile` and `Makefile` changes do break developing/building the kopf operator. Should they maybe be renamed while things are not 100% finished ?

sgaist:
@olevski I am wondering something: isn't #585 superseding this one ?

olevski:
> @olevski I am wondering something: isn't #585 superseding this one ?

@sgaist yes. But I wanted to merge this first and then #585 after that. So that the CRD changes and the scaffolding are separate PRs/commits in the history. And yeah I could just rewrite the history of #585 and rebase 2 commits onto main instead of squash and merge. But this is a bit easier for me. Plus the PR discussions between the two PRs are different and separate.

> From a code point of view I agree however the Dockerfile and Makefile changes do break developing/building the kopf operator. Should they maybe be renamed while things are not 100% finished ?

Yes let me fix these things here and then we can merge this and refine things in subsequent PRs.

olevski:
@sgaist everything should be fixed now and I adapted the devcontainer to work with the go components and also to run all the go tests in CI. So we are good to go I think. Unless I am missing something.

sgaist (approved):
LGTM !

PR #585, opened 2024-06-25
feat: AmaltheaSession CRD spec in golang

olevski:
depends on #582 

The only relevant file is `api/v1alpha1/amaltheasession_types.go`


api/v1alpha1/amaltheasession_types.go
```diff
@@ -0,0 +1,265 @@
...
+}
+
+type Culling struct {
+	// +kubebuilder:default:=0
+	// +kubebuilder:validation:Minimum:=0
+	// The maximum allowed age for a session, regardless of whether it
+	// is active or not. When the threshold is reached the session is hibernated.
+	// A value of zero indicates that Amalthea will not automatically hibernate
+	// the session based on its age.
+	MaxAgeSeconds int `json:"maxAgeSeconds,omitempty"`
```
sgaist, line 158: I am wondering whether we should use a better defined version of `int` for these type of duration field such as `int32` or `int64` since `int` can be either 32 or 64 bit depending on the architecture.

api/v1alpha1/amaltheasession_types.go
```diff
@@ -0,0 +1,265 @@
...
+	// access the session.
+	// For `oauth2proxy` please see https://oauth2-proxy.github.io/oauth2-proxy/configuration/overview#config-file.
+	SecretRef SecretRef `json:"secretRef,omitempty"`
+}
+
+// A reference to a Kubernetes secret and a specific field in the secret to be used in a session
+type SecretRef struct {
+	Name string `json:"name"`
+	Key  string `json:"key"`
+}
```
sgaist, line 205: Shouldn't we use [LocalObjectReference](https://github.com/kubernetes/api/blob/e09016fffd8ee1edc6575b8ea7c314e9c398120f/core/v1/types.go#L6647) rather than Name like
[SecretEnvSource](https://github.com/kubernetes/api/blob/e09016fffd8ee1edc6575b8ea7c314e9c398120f/core/v1/types.go#L2371) does ?
Granted LocalObjectReference currently has a single field called Name but it might be more explicit in what is expected.

api/v1alpha1/amaltheasession_types.go
```diff
@@ -0,0 +1,265 @@
...
+
+type Authentication struct {
+	Enabled bool               `json:"enabled,omitempty"`
+	Type    AuthenticationType `json:"type,omitempty"`
+	// Kubernetes secret that contains the authentication configuration
+	// For `token` generate a hard to guess string / password-like string.
+	// this value can be used as Authorization header or as a cookie with the name `amaltheaSessionToken` to
+	// access the session.
+	// For `oauth2proxy` please see https://oauth2-proxy.github.io/oauth2-proxy/configuration/overview#config-file.
+	SecretRef SecretRef `json:"secretRef,omitempty"`
```
sgaist, line 198: Looking at several objects that have SecretRef fields, it seems that all of them uses pointer for this, should we do the same ?

api/v1alpha1/amaltheasession_types.go
```diff
@@ -0,0 +1,265 @@
...
+}
+
+// A reference to a Kubernetes secret and a specific field in the secret to be used in a session
+type SecretRef struct {
```
sgaist, line 202: Shouldn't the type name be different than the field name it will be used for ? Maybe `SessionSecretRef` ?

api/v1alpha1/amaltheasession_types.go
```diff
@@ -0,0 +1,265 @@
...
+	Conditions []metav1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type" protobuf:"bytes,1,rep,name=conditions"`
+	// +kubebuilder:default:=NotReady
+	State               State           `json:"state,omitempty"`
+	URL                 string          `json:"url,omitempty"`
+	ContainerCounts     ContainerCounts `json:"readyCounts,omitempty"`
+	InitContainerCounts ContainerCounts `json:"initReadyCounts,omitempty"`
+	Idle                bool            `json:"idle,omitempty"`
+	IdleSince           metav1.Time     `json:"idleSince,omitempty"`
+	FailingSince        metav1.Time     `json:"failingSince,omitempty"`
+	HibernatedSince     metav1.Time     `json:"hibernatedSince,omitempty"`
```
sgaist, line 235: I was wondering whether we should also have an `Anonymous` field ?

api/v1alpha1/amaltheasession_types.go
```diff
@@ -0,0 +1,265 @@
...
+	// The tag, branch or commit SHA to checkout, if ommitted then will be the tip of the default branch of the repo
+	Revision string `json:"revision,omitempty"`
+	// The Kubernetes secret that contains the git configuration to be used during cloning.
+	// This can be used to inject credentials in addition to any other repo-specific Git configuration.
+	// NOTE: you have to specify the whole git config in a single key in the secret.
+	CloningGitConfigSecretRef SecretRef `json:"cloningGitConfigSecretRef,omitempty"`
+	// The Kubernetes secret that contains the git configuration to be used when the session is running.
+	// This can be used to inject credentials in addition to any other repo-specific Git configuration.
+	// NOTE: you have to specify the whole git config in a single key in the secret.
+	GitConfigSecretRef SecretRef `json:"gitConfigSecretRef,omitempty"`
```
sgaist, line 128: In the original CRD I used a dedicated Git object as I was thinking that there might be other tools that could be requested. It is unlikely that at this time code is not available through git but I do wonder if we should keep that separated so it's a field that can be swapped for something else like for example `curl` or [fossil](https://www.fossil-scm.org/home/doc/trunk/www/fossil-v-git.wiki) (note I haven't used it yet just jumped from memory).

sgaist (commented):
It's looking pretty nice, thanks !

api/v1alpha1/amaltheasession_types.go
```diff
@@ -0,0 +1,265 @@
...
+type StorageType string
+
+const Rclone StorageType = "rclone"
+
+type DataSource struct {
+	// +kubebuilder:default:=rclone
+	// The data source type
+	Type StorageType `json:"type,omitempty"`
+	// +kubebuilder:example:=data/storages
+	// +kubebuilder:default:="."
```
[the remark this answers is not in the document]
olevski: Yes that is a very good point.

api/v1alpha1/amaltheasession_types.go
olevski: K8s has a type for duration - so I used that. It will allow us to specify things like `2h5m4s`.

aledegano: AFAIK even the base time package has a `Duration` type and a function to parse a string such the one above: https://pkg.go.dev/time#ParseDuration

api/v1alpha1/amaltheasession_types.go
olevski: So I embedded `LocalObjectReference` and this is what the relevant CRD sections looks like:

```
                  secretRef:
                    description: Kubernetes secret that contains the authentication
                      configuration For `token` generate a hard to guess string /
                      password-like string. this value can be used as Authorization
                      header or as a cookie with the name `amaltheaSessionToken` to
                      access the session. For `oauth2proxy` please see https://oauth2-proxy.github.io/oauth2-proxy/configuration/overview#config-file.
                    properties:
                      key:
                        description: The secret field or key name to be used.
                        type: string
                      name:
                        description: 'Name of the referent. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
                          TODO: Add other useful fields. apiVersion, kind, uid?'
                        type: string
                    type: object
                    x-kubernetes-map-type: atomic

```

Because the docs on the `Name` field in `LocalObjectReference` are a bit weird.

I think it is better to stick with what we have above.

api/v1alpha1/amaltheasession_types.go
olevski: @sgaist what would this indicate? That the user using the session is anonymous? This can happen in 2 cases potentially - either when you dont have authentication or you have the token based authentication. And token based authentication does not guarantee that the user is anonymous. So due to this ambiguity I am not sure that we should judge this here. Someone can look at the CR and make this determination based on what their use case is.

api/v1alpha1/amaltheasession_types.go
olevski: This is a very good point. And to follow the same pattern we have elsewhere I added a type field for the code repositories. Currently the only supported value is `git` but we can expand it in the future if we need it. And it is more uniform with the data sources and authentication sections.

api/v1alpha1/amaltheasession_types.go
sgaist: Fair enough, let's keep it.

api/v1alpha1/amaltheasession_types.go
olevski: @aledegano yes and they have a type for the already in the k8s api that I can use directly and easily. So I am using that. Thanks!

api/v1alpha1/amaltheasession_types.go
```diff
@@ -0,0 +1,282 @@
...
+	// +kubebuilder:validation:Format:=date-time
+	HibernatedSince metav1.Time `json:"hibernatedSince,omitempty"`
+}
+
+//+kubebuilder:object:root=true
+//+kubebuilder:subresource:status
+
+// +kubebuilder:printcolumn:name="Status",type="string",JSONPath=`.status.state`
+// +kubebuilder:printcolumn:name="URL",type="string",JSONPath=`.status.url`
+// +kubebuilder:printcolumn:name="Ready",type="string",JSONPath=`{.status.containerCounts.ready}/{.status.containerCounts.total}`
```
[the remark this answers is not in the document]
olevski: Ah ok I did not try to actually install the CRD - which I definitely should have done. I will fix this.

api/v1alpha1/amaltheasession_types.go
```diff
@@ -0,0 +1,283 @@
...
+	// The arguments to run in the session container, if omitted it will use the Docker image CMD
+	Args []string    `json:"args,omitempty"`
+	Env  []v1.EnvVar `json:"env,omitempty"`
+	// Resource requirements and limits in the same format as a Pod in Kubernetes
+	Resources v1.ResourceRequirements `json:"resources,omitempty"`
+	// +kubebuilder:default:=8000
+	// +kubebuilder:validation:ExclusiveMinimum:=true
+	// +kubebuilder:validation:Minimum:=0
+	// The TCP port where whatever is running in the session container will listen on for connections
+	Port    int     `json:"port"`
```
sgaist, line 77: To match the ContainerPort definition:

```suggestion
	Port    int32   `json:"port"`
```

api/v1alpha1/amaltheasession_types.go
```diff
@@ -0,0 +1,283 @@
...
+	// +kubebuilder:validation:Minimum:=0
+	// The TCP port where whatever is running in the session container will listen on for connections
+	Port    int32   `json:"port"`
+	Storage Storage `json:"storage,omitempty"`
+	// The abolute path for the working directory of the session container, if omitted it will use the image
+	// working directory.
+	WorkingDir string `json:"workingDir,omitempty"`
+	// +kubebuilder:default:=1000
+	// +kubebuilder:validation:Minimum:=0
+	RunAsUser int `json:"runAsUser"`
```
sgaist, line 84: To match original type
```suggestion
	RunAsUser int64 `json:"runAsUser"`
```

api/v1alpha1/amaltheasession_types.go
```diff
@@ -0,0 +1,283 @@
...
+	// +kubebuilder:default:=1000
+	// +kubebuilder:validation:Minimum:=0
+	RunAsGroup int `json:"runAsGroup"`
```
sgaist, line 87: To match original type
```suggestion
	RunAsGroup int64 `json:"runAsGroup"`
```

sgaist (changes requested):

sgaist (approved):
LGTM !

PR #588, opened 2024-06-26
feat: add csi-rclone as dependency

aledegano:
Its installation is disabled by default.


helm-chart/amalthea/values.yaml
```diff
@@ -197,3 +199,15 @@ auditLog:
     pathPrefix: ""
     accessKeyId: ""
     secretAccessKey: ""
+
+csi-rclone:
+  install: false
+  # This section is only relevant if you are installing csi-rclone as part of Renku
+  ## Name of the csi storage class to use for RClone/Cloudstorage. Should be unique per cluster.
```
Panaetius, line 206: should we comment that this should match what's in the main renku chart?

aledegano: oh, I was planning to remove csi-rclone from renku at the same time we will include whatever release of Amalthea this goes in. WDYT?

Panaetius: no I mean this in the notebooks config needs to match the class in the csi-rclone: https://github.com/SwissDataScienceCenter/renku/blob/master/helm-chart/renku/values.yaml#L962

I think we can also make this configurable for remote cluster configuration, right now it would have to be the same in all clusters

aledegano: Ah, sorry, completely misunderstood!
Yeah, let me add a line about that!

aledegano: Changed, let me know what you think

PR #596, opened 2024-06-28
feat: implement core AmaltheaSession crd elements creation

### LINKED ISSUE #598: Core objects creation

sgaist:
## Describe your changes

This PR implements the creation part of the core elements of the new AmaltheaSession CRD.

This includes:

- StatefulSet
- Matching Service
- Matching Ingress

Close #598 


olevski (changes requested):
Thanks @sgaist this looks really good. 

But lets do the following refactoring to keep the reconcile loop a bit cleaner.

Create a struct that can hold all of Amaltheas children resources. I.e.

```go
type AmaltheaChildren struct {
	Ingress *networkingv1.Ingress
	Service v1.Service
	StatefulSet appsv1.StatefulSet
	PVC v1.PersistentVolumeClaim
}
```

Then we have a function on the CR to generate this from the manifest:

```go
func (cr *AmaltheaSession) Children() AmaltheaChildren {
    ...
}
```

Then we can use the output of this function to diff the current state against what is desired based on the spec. This function would always return what is desired based on the spec.

internal/controller/amaltheasession_controller.go
```diff
@@ -47,16 +66,338 @@ type AmaltheaSessionReconciler struct {
...
+				},
+			},
+		},
+	}
+
+	// Set the ownerRef for the StatefulSet
+	// More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/
+	if err := ctrl.SetControllerReference(amaltheasession, sts, r.Scheme); err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
sgaist: Done

api/v1alpha1/amaltheasession_types.go
```diff
@@ -269,6 +275,184 @@ type AmaltheaSession struct {
...
+func (cr *AmaltheaSession) Children() AmaltheaChildren {
+	return AmaltheaChildren{
+		StatefulSet: statefulSetForAmaltheaSession(cr),
+		Service:     serviceForAmaltheaSession(cr),
+		Ingress:     ingressForAmaltheaSession(cr),
+	}
+}
+
+// statefulSetForAmaltheaSession returns a AmaltheaSession StatefulSet object
+func statefulSetForAmaltheaSession(
```
sgaist, line 306: @olevski Should we also make these free functions methods of `AmaltheaSession` struct ?

One reason would be to make their call more intuitive when outside of this file when not needing all of the children for example when checking the pod created by the `StatefulSet`

olevski: Yes that would be really nice! After that I will approve and we can merge.

sgaist: Done !

olevski (approved):

PR #604, opened 2024-07-02
feature: crd fields specification cleanup

### LINKED ISSUE #602: Define immutable fields

### LINKED ISSUE #603: Define which field should be allowed to be nil (absence VS default null value).

sgaist:
## Describe your changes

This PR refactors the qualifiers/validation of the fields used in various parts of the new AmaltheaSession CRD.

The goal is to clarify which are optional as in "absent" allowed and which are mandatory

Close #602
Close #603 


api/v1alpha1/amaltheasession_types.go
```diff
@@ -99,15 +97,17 @@ type Session struct {
...
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="PathPrefix is immutable"
+	PathPrefix string `json:"pathPrefix,omitempty"`
 	// The name of the TLS secret, same as what is specified in a regular Kubernetes Ingress.
 	TLSSecretName string `json:"tlsSecretName,omitempty"`
 }
 
 type Storage struct {
 	ClassName string            `json:"storageClassName,omitempty"`
-	Size      resource.Quantity `json:"storageSize,omitempty"`
+	Size      resource.Quantity `json:"storageSize"`
```
olevski, line 110: Hey so in my other PR #605 I made size default to `1Gi`. Is that OK? If someone is just trying to get things going it is nicer to require fewer defaults in the CR spec.

api/v1alpha1/amaltheasession_types.go
```diff
@@ -207,14 +207,15 @@ const Token AuthenticationType = "token"
 const Oidc AuthenticationType = "oauth2proxy"
 
 type Authentication struct {
+	// +kubebuilder:default:=true
```
olevski, line 210: Check #608 and how size/Storage is handled I think you need default:={} at the Spec level on the Authentication field so that this default really shows up.

olevski (changes requested):

api/v1alpha1/amaltheasession_types.go
sgaist: Sure thing !

There might be some things that we need to think about in the light of [Optional VS Required](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required) from the API conventions.

olevski (approved):
Thanks @sgaist!

PR #605, opened 2024-07-02
feat: basic status and reconcile implementation

olevski:
@sgaist 

I did a few changes from your version that was just merged:
- moved the methods that generate the children in their own file separate from the CRD spec
- used the `controllerutil` library to simplify reconciliation
- removed the status condition changes - I am not quite sure how to handle these yet and I did not want to have them in the main reconcile function
- moved a the reconcile code to separate methods so that the top most reconcile loop code is shorter
- added PVCs
- implemented hibernation

A few things that are still missing, that I will do in a followup PR:
- the ready/total containers on the pod in the status are still missing
- the ingress is still optional and the url in the status gets generated incorrectly


api/v1alpha1/amaltheasession_children.go
```diff
@@ -0,0 +1,220 @@
...
+	if cr.Spec.Hibernated {
+		replicas = 0
+	}
+
+	session := cr.Spec.Session
+	pvc := cr.PVC()
+
+	sessionContainer := v1.Container{
+		Image:           session.Image,
+		Name:            "session",
```
sgaist, line 31: Now I remember I wanted to ask something for this, should we have a constant for this ?

api/v1alpha1/amaltheasession_children.go
```diff
@@ -0,0 +1,220 @@
...
+		ImagePullPolicy: v1.PullIfNotPresent,
+
+		Ports: []v1.ContainerPort{{
+			ContainerPort: session.Port,
+			Name:          "session-port",
```
sgaist, line 36: Same question as above, should we use a constant and maybe it's content be dependent on the sessionContainer name.

api/v1alpha1/amaltheasession_children.go
```diff
@@ -0,0 +1,220 @@
...
+			Protocol:      v1.ProtocolTCP,
+		}},
+
+		Args:                     session.Args,
+		Command:                  session.Command,
+		Env:                      session.Env,
+		Resources:                session.Resources,
+		VolumeMounts:             []v1.VolumeMount{{Name: "session", MountPath: session.Storage.MountPath}},
```
sgaist, line 44: Same constant question

api/v1alpha1/amaltheasession_children.go
```diff
@@ -0,0 +1,220 @@
...
+				MatchLabels: labels,
+			},
+			Template: v1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: labels,
+				},
+				Spec: v1.PodSpec{
+					Containers:     containers,
+					InitContainers: cr.Spec.ExtraInitContainers,
+					Volumes:        []v1.Volume{{Name: "session", VolumeSource: v1.VolumeSource{PersistentVolumeClaim: &v1.PersistentVolumeClaimVolumeSource{ClaimName: pvc.Name}}}},
```
sgaist, line 82: Two things here:
- Again, the question of constant for the Name field
- Formatting wise, it's the only place that packs the structure on one line

Also, are we sure that we will always have a PVC ? Thinking about the case of anonymous session and also the ephemeral use cases (BinderHub like).

api/v1alpha1/amaltheasession_children.go
```diff
@@ -0,0 +1,220 @@
...
+
+	ingress := cr.Spec.Ingress
+	if reflect.DeepEqual(ingress, Ingress{}) {
+		return nil
+	}
+
+	path := "/"
+	if ingress.PathPrefix != "" {
+		path = ingress.PathPrefix
+	}
```
sgaist, line 125: I am just realizing, shouldn't we rather have PathPrefix defined with a default at "/" ?

api/v1alpha1/amaltheasession_children.go
```diff
@@ -0,0 +1,220 @@
...
+				SecretName: ingress.TLSSecretName,
+			}},
+		},
+	}
+
+	return ing
+}
+
+// PVC returned the desired specification for a persistent volume claim
+func (cr *AmaltheaSession) PVC() v1.PersistentVolumeClaim {
```
sgaist, line 170: Isn't the Storage optional ?

sgaist (commented):
Thanks for the refactoring !

This is looking nice

api/v1alpha1/amaltheasession_children.go
olevski: Yes absolutely. I will add that.

api/v1alpha1/amaltheasession_children.go
olevski: I will do this in a followup PR. It should be and as you mentioned above we should allow people to use emptyDir or ephemeral volumes or something similar.

api/v1alpha1/amaltheasession_children.go
olevski: done

olevski:
@sgaist I spent half the day trying to figure out why the owner references were failing in the tests. So I went back to what you had before and what I made you change. We should have kept that since it is what the k8s docs recommend. My bad about that.

With the SetControllerReference (as you had it earlier) things do not fail. They way I proposed was fragile because sometimes the kind/apiversion can be empty and the SetControllerReference handles this gracefully.

api/v1alpha1/amaltheasession_children.go
```diff
@@ -0,0 +1,208 @@
...
+									Port: networkingv1.ServiceBackendPort{
+										Name: servicePortName,
+									},
+								},
+							},
+						}},
+					},
+				},
+			}},
+			TLS: []networkingv1.IngressTLS{{
```
sgaist, line 159: Shouldn't the TLS part depend on the presence of TLSSecretName ?

sgaist (commented):
That OwnerReferences issue was a sneaky one !

Other than my question related to TLS, things are looking good for me.

api/v1alpha1/amaltheasession_children.go
olevski: Yes definitely. And to follow the guide on optional vs requried values in CRD I marked that as optional and converted it to a pointer.

sgaist (approved):
LGTM !

PR #609, opened 2024-07-02
feat: additional status updates for AmaltheaSessions

### LINKED ISSUE #608: Session status update

olevski:
closes #608 

The only thing missing from the status now is the idleness check. But we should resolve that in the hibernation task.

Depends on #605


internal/controller/children.go
```diff
@@ -167,20 +168,24 @@ func (c ChildResourceUpdates) AllEqual(op controllerutil.OperationResult) bool {
 	return ingressOK && c.Service.UpdateResult == op && c.PVC.UpdateResult == op && c.StatefulSet.UpdateResult == op
 }
 
-func (c ChildResourceUpdates) IsRunning(cr *amaltheadevv1alpha1.AmaltheaSession) bool {
+func (c ChildResourceUpdates) IsRunning(pod *v1.Pod) bool {
 	onlyStatusUpdates := c.AllEqual(controllerutil.OperationResultUpdatedStatusOnly)
 	noUpdates := c.AllEqual(controllerutil.OperationResultNone)
 	ssReady := c.StatefulSet.Manifest.Status.ReadyReplicas == 1 && c.StatefulSet.Manifest.Status.Replicas == 1
```
sgaist, line 174: I missed it before but I would suggest to use the k8s short name so it makes parsing easier ( i.e. `stsReady`).

internal/controller/children.go
```diff
@@ -167,20 +168,24 @@ func (c ChildResourceUpdates) AllEqual(op controllerutil.OperationResult) bool {
...
-	return ssReady && (onlyStatusUpdates || noUpdates)
+	podExists := pod != nil
+	podReady := podExists && podIsReady(pod)
+	return ssReady && podReady && (onlyStatusUpdates || noUpdates)
```
sgaist, line 177: With regard to my suggestion above
```suggestion
	podExists := pod != nil
	podReady := podExists && podIsReady(pod)
	return stsReady && podReady && (onlyStatusUpdates || noUpdates)
```

internal/controller/status.go
```diff
@@ -5,25 +5,60 @@ import (
...
+	initCounts, counts := containerCounts(pod)
+	return initCounts.Ok() && counts.Ok() && phaseOk
+}
+
+func podIsFailed(pod *v1.Pod) bool {
+	if pod == nil || pod.GetDeletionTimestamp() != nil {
+		// A missing pod or a pod being deleted is not considered failed
+		return false
+	}
+	restartThreshold := int32(3)
```
sgaist, line 46: Shouldn't we have that as a constant ?
Is it something that we want to be configurable ? If yes, at which level ? Controller or CRD ?

sgaist (changes requested):

internal/controller/status.go
olevski: Yes I will move that as a constant. 

But sometime in the future we should have that as part of the CR. Because unfortunately k8s does not natively support such a concept. I.e. you can set your pod to either always or never restart when it fails. Sure it has exponential (or some kind of a backoff) but still once a session is failing it will keep restarting forever.

sgaist (approved):
LGTM

PR #618, opened 2024-07-04
feat: expand the CR spec with additional vols

### LINKED ISSUE #617: Enable specifying additional volumes

olevski:
This also makes the port field more general so that it can point to any port on the session Pod rather than only to the session container.

This gives more flexibility when using the extra containers field to make the ingress route to the extra containers (if required) rather than only to the session container.

closes #617


sgaist (approved):
Just one question: did we discuss about having the storage optional as in possibly absent ?

olevski:
> Just one question: did we discuss about having the storage optional as in possibly absent ?

No not really but I think we can tackle this in a followup either together or separate with the ephemeral storage.

sgaist:
> > Just one question: did we discuss about having the storage optional as in possibly absent ?
> 
> No not really but I think we can tackle this in a followup either together or separate with the ephemeral storage.

Found it :-D
It's a floating question: https://github.com/SwissDataScienceCenter/amalthea/pull/605#discussion_r1664017096

I remember writing somewhere that it would be useful for anonymous and/or BinderHub like use cases but it can be part of the ephemeral storage task.

PR #628, opened 2024-07-11
AmaltheaSession culling

### LINKED ISSUE #616: Culling

SalimKayal:
This merge requests adds the ability to automatically scale down the AmaltheaSession's Stateful Set in case of failure or inactivity. After `MaxHibernatedDuration` it deletes the custom resource.

closes #616


internal/controller/cull.go
```diff
@@ -0,0 +1,81 @@
...
+	"context"
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	amaltheadevv1alpha1 "github.com/SwissDataScienceCenter/amalthea/api/v1alpha1"
+)
+
+func customResourceNeedsDeletion(amaltheasession *amaltheadevv1alpha1.AmaltheaSession) bool {
```
sgaist, line 29: This method should be part of the interface of the AmaltheaSession struct since its using only information from that struct.

internal/controller/cull.go
```diff
@@ -0,0 +1,81 @@
...
+	culling := amaltheasession.Spec.Culling
+	status := amaltheasession.Status
+	now := time.Now()
+	hibernatedDuration := now.Sub(status.HibernatedSince.Time)
+	return status.State == amaltheadevv1alpha1.Hibernated &&
+		hibernatedDuration > culling.MaxHibernatedDuration.Duration
+}
+
+func updateStatefulSetReplicas(ctx context.Context, r *AmaltheaSessionReconciler, amaltheasession *amaltheadevv1alpha1.AmaltheaSession) error {
```
sgaist, line 38: Following discussion the name of the function is not correct anymore:

```suggestion
func updateHibernationState(ctx context.Context, r *AmaltheaSessionReconciler, amaltheasession *amaltheadevv1alpha1.AmaltheaSession) error {
```

sgaist (changes requested):

internal/controller/amaltheasession_controller.go
```diff
@@ -120,6 +120,17 @@ func (r *AmaltheaSessionReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		}
 	}
 
+	if customResourceNeedsDeletion(amaltheasession) {
+		err = r.Client.Delete(ctx, amaltheasession)
+		log.Info("custom resource deleted")
+		return ctrl.Result{}, err
+	}
+
+	err = updateStatefulSetReplicas(ctx, r, amaltheasession)
```
SalimKayal, line 129: Then this has to be changed too.

```suggestion
	err = updateHibernationState(ctx, r, amaltheasession)
```

internal/controller/cull.go
SalimKayal: Done

sgaist (approved):
LGTM

PR #631, opened 2024-07-11
feat: implement shm support

### LINKED ISSUE #630: Add support for adjusting /dev/shm
For adjusting `/dev/shm` this is how we patch it in the old sessions:

```
            {
                "type": "application/json-patch+json",
                "patch": [
                    {
                        "op": "add",
                        "path": "/statefulset/spec/template/spec/volumes/-",
                        "value": {
                            "name": "shm",
                            "emptyDir": {
                                "medium": "Memory",
                                # NOTE: We are giving /dev/shm up to half of the memory request
                                "sizeLimit": int(server.server_options.storage / 2),
                            },
                        },
                    },
                    {
                        "op": "add",
                        "path": "/statefulset/spec/template/spec/containers/1/volumeMounts/-",
                        "value": {
                            "mountPath": "/dev/shm",
                            "name": "shm",
                        },
                    },
                ],
            }
```

So it is a memory backed volume mounted at `/dev/shm`. And its size is half the memory request for the session. Here we can make it part of the CRD with a default value of 1Gi or something like that.

_Originally posted by @olevski in https://github.com/SwissDataScienceCenter/amalthea/issues/597#issuecomment-2213228409_

sgaist:
## Describe your changes

This PR implements two things:

- Adding a /dev/shm volume to the StatefulSet
- The option to adjust /dev/shm in the CRD

Closes #630

### DIFF
--- a/api/v1alpha1/amaltheasession_children.go
+++ b/api/v1alpha1/amaltheasession_children.go
@@ -45,13 +45,10 @@ func (cr *AmaltheaSession) StatefulSet() appsv1.StatefulSet {
 				Name:      sessionVolumeName,
 				MountPath: session.Storage.MountPath,
 			},
-			{
-				Name:      shmVolumeName,
-				MountPath: "/dev/shm",
-			},
 		},
 		extraMounts...,
 	)
+
 	volumes := []v1.Volume{
 		{
 			Name: sessionVolumeName,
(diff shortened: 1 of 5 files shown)


SalimKayal:
Do we want that to be mandatory or an option ?

I get that this was always mounted in the previous implementation though I am not sure this is always needed. Any inputs ?

api/v1alpha1/amaltheasession_types.go
```diff
@@ -89,6 +89,10 @@ type Session struct {
 	// +optional
 	// +kubebuilder:default:={}
 	Storage Storage `json:"storage,omitempty"`
+	// +optional
+	// +kubebuilder:default:="500Mi"
+	// Size of /dev/shm
+	ShmSize resource.Quantity `json:"shmSize,omitempty"`
```
olevski, line 95: I thought the recommendation was that optional values are pointers. I know we dont follow that everywhere but we can start. I wish I had seen that api convention doc a bit earlier from when we started working on the new operator.

https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required

api/v1alpha1/amaltheasession_children.go
```diff
@@ -22,6 +22,7 @@ const sessionContainerName string = prefix + "session"
 const servicePortName string = prefix + "http"
 const servicePort int32 = 80
 const sessionVolumeName string = prefix + "volume"
+const shmVolumeName string = "dev-shm"
```
olevski, line 25:
```suggestion
const shmVolumeName string = prefix + "dev-shm"
```

olevski (changes requested):
Just 2 tiny changes. Otherwise it looks good!

olevski:
> Do we want that to be mandatory or an option ?

> I get that this was always mounted in the previous implementation though I am not sure this is always needed. Any inputs ?

Actually I think @SalimKayal is right, it is probably better to not always have this. I know people want this when they use tensorflow/pytorch but whether they always need it or only when they are doing something fancier like using multiple gpus I cannot remember.

@sgaist what do you think?

api/v1alpha1/amaltheasession_types.go
sgaist: I had in mind that SHM would be a "mandatory" item hence the use of a value. But following the discussion below it makes even more sense to have it as a pointer.

I'll update that.

sgaist:
@olevski I agree as well, it will also make the session lighter on resource.

olevski (approved):
Thanks this looks good. But there is a merge conflict. So let me know and I will re-approve when you resolve it.

sgaist:
> Thanks this looks good. But there is a merge conflict. So let me know and I will re-approve when you resolve it.

Done, the tests were conflicting :-)

olevski (approved):

PR #644, opened 2024-07-19
feat: implement authentication through oauth2proxy

### LINKED ISSUE #614: Authentication for registered users
In the old amalthea we use oauth2proxy but here we can also use something else but with similar functionality, it does not have to be oauth2proxy.

See here for how to configure oauth2 proxy: https://oauth2-proxy.github.io/oauth2-proxy/configuration/overview/#config-file

It is simplest if we expect the configuration to all be in a single key in the secret that users can reference in the CR.

sgaist:
## Describe your changes

This PR implements authentication for registered users in a similar fashion as Amalthea's v1 using oauth2-proxy.

As a drive-by fix, it modifies the culling handling to avoid re-triggering the reconcile loop if the pod is missing.

Closes #614 

### DIFF
--- a/api/v1alpha1/amaltheasession_children.go
+++ b/api/v1alpha1/amaltheasession_children.go
@@ -12,6 +12,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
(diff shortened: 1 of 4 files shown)


api/v1alpha1/amaltheasession_children.go
```diff
@@ -101,20 +102,53 @@ func (cr *AmaltheaSession) StatefulSet() appsv1.StatefulSet {
...
+				VolumeMounts: []v1.VolumeMount{
+					{
+						Name:      "proxy-configuration-secret",
+						MountPath: "/etc/oauth2-proxy",
+					},
+				},
+			}
+			containers = append(containers, authContainer)
+		}
+	}
```
SalimKayal, line 150: Maybe for the sake of concisenenss

```suggestion
	if auth := cr.Spec.Authentication; auth != nil && auth.Enabled && auth.Type == Oidc {

		volumes = append(volumes, v1.Volume{
			Name: "proxy-configuration-secret",
			VolumeSource: v1.VolumeSource{
				Secret: &v1.SecretVolumeSource{
					SecretName: auth.SecretRef.Name,
					Optional:   ptr.To(false),
				},
			},
		})

		authContainer := v1.Container{
			Image: "bitnami/oauth2-proxy:7.4.0",
			Name:  "oauth2-proxy",
			SecurityContext: &v1.SecurityContext{
				AllowPrivilegeEscalation: ptr.To(false),
				RunAsNonRoot:             ptr.To(true),
			},
			Args: []string{"--config=/etc/oauth2-proxy/" + auth.SecretRef.Key},
			VolumeMounts: []v1.VolumeMount{
				{
					Name:      "proxy-configuration-secret",
					MountPath: "/etc/oauth2-proxy",
				},
			},
		}
		containers = append(containers, authContainer)
	}
```

api/v1alpha1/amaltheasession_children.go
```diff
@@ -284,5 +318,14 @@ func (cr *AmaltheaSession) AllSecrets() v1.SecretList {
...
 	}
 
+	if cr.Spec.Authentication != nil && cr.Spec.Authentication.Enabled {
+		secrets.Items = append(secrets.Items, v1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				Namespace: cr.Namespace,
+				Name:      cr.Spec.Authentication.SecretRef.Name,
+			},
+		})
+	}
```
SalimKayal, line 328: same as above

```suggestion
	if auth := cr.Spec.Authentication; auth != nil && auth.Enabled {
		secrets.Items = append(secrets.Items, v1.Secret{
			ObjectMeta: metav1.ObjectMeta{
				Namespace: cr.Namespace,
				Name:      auth.SecretRef.Name,
			},
		})
	}
```

SalimKayal (commented):
but any comments below are debatable. otherwise LGTM.

api/v1alpha1/amaltheasession_children.go
SalimKayal, line 1: on an other topic, the StatefulSet method is starting to look too big to me, perhaps we should split every specific part of it to enhance readability.

SalimKayal: but this can (and probably should) be done in an other PR

PR #646, opened 2024-07-19
Add status conditions

### LINKED ISSUE #642: Add status conditions
Expose status conditions

At first expose trivial conditions so that when needed information is pinpointed it becomes easy to add.

SalimKayal:
This adds trivially available status conditions to the `AmaltheaSession` custom resource.

closes #642


olevski (changes requested):
I am sorry I got really picky in this review. But I think it is good to remove the nesting in the IF logic. And the one-liner if-statement + assignment pattern in go really makes it hard to glance and understand the logic in a series of if statement branches like what we have here. So I think it is nicer to have more lines of code but logic that you can parse a bit easier.

olevski (approved):
Thanks Salim!

PR #668, opened 2024-08-06
fix: properly pass down namespace configuration

olevski:
The way that the controller is started was changed. But the ingesting of configuration values for which namespaces should be watched was not fully updated.


olevski:
This version is deployed here: https://github.com/SwissDataScienceCenter/renku/pull/3727 and all tests are passing.

aledegano (approved):

PR #674, opened 2024-08-08
feat: extra volumes in authentication container

olevski:
This allows for someone to use a file for a list of emails that would be allowed to access the session with oauth2 proxy. The change only applies to amaltheasesssions.

Note that following files are autogenerated:
- config/crd/bases/amalthea.dev_amaltheasessions.yaml
- helm-chart/amalthea-sessions/templates/amaltheasession-crd.yaml
- api/v1alpha1/zz_generated.deepcopy.go


SalimKayal (approved):
LGTM

PR #677, opened 2024-08-10
fix: constrain authproxy setup

olevski:
The proxy setup was too flexible and I think this can easily lead to issues where users think that they know what they are doing but are breaking things.

The operator is well suited to know and decide exactly where things are getting routed. So the configuration of these things should be in control of the users of amalthea sessions.

Also I simplified the path prefix and now there is only one - which is specified in the session section. With the 2 we had before one in ingress and another one in the session section things were confusing.


olevski:
Also I hardcoded the port that will be used by authproxy and removed that from the valid values for ports that the user can indicate in the spec. This should also prevent weird edge cases in the future.

olevski:

- #677 ◀
    - #678

authproxy/cmd/serve.go
```diff
@@ -107,10 +97,22 @@ func serve(cmd *cobra.Command, args []string) {
...
 		e.Logger.Fatal(err)
 	}
+	proxyPortStr := os.Getenv(authProxyPortEnvKey)
+	if proxyPortStr == "" {
+		proxyPortStr = "65535"
+	}
+	authProxyPort, err := strconv.Atoi(proxyPortStr)
+	if err != nil {
+		e.Logger.Fatalf("Cannot convert the proxy port environment variable to integer: %v", err)
+	}
```
sgaist, line 115: I am not sure I am following the why of this change.

Both the port and URL could already be passed through environment variables with the original implementation.

See the two BindEnv calls you removed. Also, in `root.go`, I set the prefix for the environment variables to `AUTHPROXY` so you can use `AUTHPROXY_REMOTE` and `AUTHPROXY_PORT`.

olevski: @sgaist the main reason for this is that I do not want to allow people to modify the port and remote for the proxy. Because in order to do this effectively you need to know a lot of intricacies about how amalthea sessions are setup. And it is easy to misunderstand this or mess it up. 

Also IMO allowing this to be modifiable leaks implementation details that should be hidden behind the CRD. The new CRD gives 3 options when it comes to authentication which are already really flexible. I.e. no authentication, authentication via oauth2 or authentication with a cookie/header. If someone wants to do something fancier they can turn off the amalthea authentication, add their own fully custom containers and add their own custom ingress. But when users go for the authentication amalthea provides, we constrain things a little bit but we remove the guesswork of what the remote url and port should be in the oauth proxy configuration. We can already put this configuration together from the things that a user would provide in the CR.

sgaist: OK, I think I am starting to follow your trail of thoughts. I had something similar in mind and the current implementation already provides the behaviour you want based on the order of handling used by viper as described [here](https://github.com/spf13/viper#why-viper):
```
Viper uses the following precedence order. Each item takes precedence over the item below it:

    explicit call to Set
    flag
    env
    config
    key/value store
    default
```

Environment variables go above configuration file so even if a user tries to change a value there, the final say is in the operator setting the environment variables or the corresponding parameter. So we have best of both worlds: the command is still easily testable while the configuration can be overridden by the operator when necessary.
Note that not all settings currently have matching environment variables.

olevski: Ok yeah it was not clear to me how viper loads configurations. That makes sense. So I put the flags about port and url I removed. And I set them in the children resources via environment variables.

sgaist (approved):

authproxy/cmd/serve.go
```diff
@@ -18,6 +18,8 @@ package cmd
 
 import (
 	"fmt"
+	"os"
+	"strconv"
```
sgaist, line 22: I saw the lines and missed them afterward:
```suggestion
```

sgaist (changes requested):

sgaist (approved):

PR #703, opened 2024-08-23
feat(cloner): clone strategies

sgaist:
## Describe your changes

This PR implements strategies about acting before cloning.

There are three strategies:
- default: do as usual and git clone
- notifexists: don't clone if the target folder already exists
- overwrite: delete the target folder before cloning

This happens before trying to clone anything.

The goal is to allow restarting the session which means also the init container and not fail the cloning if it already happened before.

As a drive by cleanup, the `.git` extension is removed if part of the URL as happen when using the git command.

Part of: #612

### DIFF
--- a/cloner/cmd/clone.go
+++ b/cloner/cmd/clone.go
@@ -40,14 +40,22 @@ const PathFlag string = "path"
 const VerboseFlag string = "verbose"
 const StrategyFlag string = "strategy"
 
+// Pre cloning strategies
+
+const NotIfExist string = "notifexist" // Do not clone if target already exists
+const Overwrite string = "overwrite"   // Remove target first if it exists
+const NoStrategy string = "nostrategy" // Let git handle the situation
+
 var (
-	configPath       string
-	remote           string
-	revision         string
-	path             string
-	verbose          bool
-	CloningStratgies []string = []string{"notifexist", "overwrite", "default"}
-	strategy                  = newEnum(CloningStratgies, "default")
+	configPath           string
+	remote               string
+	revision             string
+	path                 string
+	verbose              bool
+	PreCloningStrategies []string = []string{
+		NotIfExist, Overwrite, NoStrategy,
+	}
+	preCloningStrategy = newEnum(PreCloningStrategies, NoStrategy)
 )
 
 type CloneFonfig struct {
@@ -71,7 +79,7 @@ func init() {
 
 	cloneCmd.Flags().BoolVar(&verbose, VerboseFlag, false, "make the command verbose")
 
-	cloneCmd.Flags().VarP(strategy, StrategyFlag, "", "the cloning strategy")
+	cloneCmd.Flags().VarP(preCloningStrategy, StrategyFlag, "", "the pre cloning strategy")
 }
 
 var cloneCmd = &cobra.Command{
@@ -99,7 +107,7 @@ func clone(cmd *cobra.Command, args []string) {
 		clonePath = path + "/" + projectName
 	}
 
-	if !strategy.Equal("default") {
+	if !preCloningStrategy.Equal(NoStrategy) {
 		// check if target exists
 		_, err := os.Stat(clonePath)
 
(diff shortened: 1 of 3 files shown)


olevski (changes requested):
One question about the "default" strategy. All else looks good.

cloner/cmd/clone.go
```diff
@@ -32,18 +33,21 @@ import (
...
-	remote     string
-	revision   string
-	path       string
-	verbose    bool
+	configPath       string
+	remote           string
+	revision         string
+	path             string
+	verbose          bool
+	CloningStratgies []string = []string{"notifexist", "overwrite", "default"}
```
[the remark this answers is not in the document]
sgaist: The `default` value means "no pre-processing" before cloning so if the target folder exists, the clone command will abort.

The other two will trigger checks and actions before the cloning happens.

cloner/cmd/clone.go
```diff
@@ -78,20 +84,44 @@ func clone(cmd *cobra.Command, args []string) {
 
 	endpoint, err := transport.NewEndpoint(remote)
 	if err != nil {
-		log.Fatal("failed to parse remote", err)
+		log.Fatal("failed to parse remote: ", err)
 	}
 
 	splittedRepo := strings.FieldsFunc(endpoint.Path, func(c rune) bool { return c == '/' }) // FieldsFunc handles repeated and beginning/ending separator characters more sanely than Split
 	if len(splittedRepo) < 2 {
```
leafty, line 91: Why does this rule exist? Just check if there is a last part, i.e. `len() > 0`.

sgaist: The standard repo URL for providers such as GitHub and GitLab have at least two components:
- namespace or group name (e.g. username, enterprise name, etc.)
- project name

for GitLab, you can also have subgroups.
This check avoids trying to clone wrongly formatted URL. Less ressources and time are wasted.

leafty: But what if I want to clone a self-hosted git repository at `https://example.org/my-repo.git`? This check seems restrictive AND superfluous. If the URL is wrong, then `git clone` will fail, there is no benefit from not allowing it to fail.

sgaist: I see you point.

The check has been updated

cloner/cmd/clone.go
```diff
@@ -78,20 +84,44 @@ func clone(cmd *cobra.Command, args []string) {
...
 	}
 
+	if !strategy.Equal("default") {
+		_, err := os.Stat(clonePath)
+
+		if err != nil && !errors.Is(err, os.ErrNotExist) {
+			log.Fatal("unexpected error: ", err)
+		}
+
+		if err == nil {
```
aledegano, line 109: [I am out of my turf, but I am interested nonetheless]
This check is not necessary no? If `err != nil` we already exited with RC 1 in the block above. For readability I would rather remove this `if`.

cloner/cmd/clone.go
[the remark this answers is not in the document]
sgaist: I have refactored this part a bit to make things clearer with better naming and additional comments.

cloner/cmd/clone.go
aledegano: Right, sorry I got confused (and that is my point): too many nesting levels and negations for me to follow.

I think that I would rewrite this block as following:
```go
_, err := os.Stat(clonePath)

if err != nil && !errors.Is(err, os.ErrNotExist) {
  log.Fatal("unexpected error: ", err)
}

if errors.Is(err, os.ErrNotExist) {
  clonePathExists := true
}

if preCloningStrategy.Equal(NotIfExist) && clonePathExists {
  log.Print(clonePath, " already exist, doing nothing")
  os.Exit(0)
}

if preCloningStrategy.Equal(Overwrite)  && clonePathExists {
  log.Print("deleting clone")
  err = os.RemoveAll(clonePath + "/")
  if err != nil {
	log.Fatal("failed to remove existing clone: ", err)
  }
}
 
# Bonus log message, just for clarity
if preCloningStrategy.Equal(NoStrategy)  && clonePathExists {
 log.Print(clonePath, " already exists, but no pre-cloning strategy selected: git will handle this.")
}
```

Would you find this more readable? I would prefer it, but I will let you decide.

sgaist: Your code is wrong on several places:
```
if errors.Is(err, os.ErrNotExist) {
  clonePathExists := true
}
```
should rather be: 
```
if errors.Is(err, os.ErrNotExist) {
  clonePathExists := false
}
```
but even so, clonePathExists lifetime is only within this if block so you will have two errors:
- Unused variable within that if block
- Undefined variable outside that if block

For the bonus log message, whether clonePathExists is true or false does not matter, the goal of this one is to just do nothing.

Anyway, your suggestion gave me a new idea on the implementation that I think should be satisfactory.

aledegano: Yes, sorry the code was sort of a braindump. But I see that you understood what I meant and I find the new implementation much much easier to follow!
Thank you!

olevski (approved):

PR #714, opened 2024-09-01
feat: specify secret adoption for every secret

olevski:

PR #704, opened 2024-08-23
fix: requeue without delay if the status is changing

olevski:
The status for amalthea sessions takes a long time to show up. This should address this problem.

PR #715, opened 2024-09-01
feat: add cloud storage

olevski:


olevski:

- #704
    - #714
        - #715 ◀

internal/controller/children.go
```diff
@@ -160,21 +174,30 @@ func (c ChildResources) Reconcile(ctx context.Context, clnt client.Client, cr *a
...
+		dataSourcesOK = dataSourcesOK && (ds.UpdateResult == op)
+	}
+	return ingressOK && c.Service.UpdateResult == op && c.PVC.UpdateResult == op && c.StatefulSet.UpdateResult == op && dataSourcesOK
 }
 
 func (c ChildResourceUpdates) IsRunning(pod *v1.Pod) bool {
-	onlyStatusUpdates := c.AllEqual(controllerutil.OperationResultUpdatedStatusOnly)
-	noUpdates := c.AllEqual(controllerutil.OperationResultNone)
+	// onlyStatusUpdates := c.AllEqual(controllerutil.OperationResultUpdatedStatusOnly)
+	// noUpdates := c.AllEqual(controllerutil.OperationResultNone)
```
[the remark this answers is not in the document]
olevski: Yes that makes sense. I think ultimately we will end up removing that. But for now I added a TODO and a clarification of why the checks were removed.

internal/controller/children.go
```diff
@@ -160,21 +174,34 @@ func (c ChildResources) Reconcile(ctx context.Context, clnt client.Client, cr *a
...
+	}
+	return ingressOK && c.Service.UpdateResult == op && c.PVC.UpdateResult == op && c.StatefulSet.UpdateResult == op && dataSourcesOK
 }
 
 func (c ChildResourceUpdates) IsRunning(pod *v1.Pod) bool {
-	onlyStatusUpdates := c.AllEqual(controllerutil.OperationResultUpdatedStatusOnly)
-	noUpdates := c.AllEqual(controllerutil.OperationResultNone)
+    // TODO: Try to re-enable the two checks below and potentially use them to determine readiness.
+	// Currently the resources created by the operator have slight changes that k8s itself applies in a few places outside
+    // of the status field. So these are picked up by the functions below. For example a PVC or a statefulset gets automatic
```
sgaist, line 197: Just to keep things neatly aligned
```suggestion
	// TODO: Try to re-enable the two checks below and potentially use them to determine readiness.
	// Currently the resources created by the operator have slight changes that k8s itself applies in a few places outside
	// of the status field. So these are picked up by the functions below. For example a PVC or a statefulset gets automatic
```

sgaist (changes requested):

leafty (commented):
FYI, the feature does not seem to work when more than one remote data storage location should be mounted.

When starting with 2 storage ids, the PVC for the first storage is missing:
```
kubectl -n renku-ci-ds-418 get pvc

7c44436c-f3-renku-2-d501bf9221e6cf7f024db        Bound    pvc-682c0e2a-2a3c-4be2-ba2a-76089741f215   1Gi        RWO            csi-cinder-sc-delete           2m29s
n7c44436c-f3-renku-2-d501bf9221e6cf7f024db-ds-1   Bound    pvc-7d665125-b549-4f1e-8c18-e6259766b46f   1Gi        ROX            csi-rclone-secret-annotation   2m28s
```

The one for `-ds-0` is missing.

Tested on this deployment -> https://renku-ci-ds-418.dev.renku.ch/

olevski:
Thanks @leafty I will take a look.

olevski:
@leafty so I was taking a pointer of a loop variable in the go code. And this will always work how you dont expect it. Because the loop variable is the same for the whole loop duration -not for each loop iteration. So if you have a loop variable and then you take its pointer and store it in a list and use that after the loop has finished what you have is a list of the same variable - i.e. the last value that you looped through.

This is fixed in go 1.22 but the operator framework is still on 1.21 and I did not want to force an update. There is a super easy fix for this in the code. But it gets me every single time.

Now you see both pvcs for the datasets:
image

The mounting still fails though because I need #737 so that I can run the sessions on the nodes dedicated for user sessions. On the dev cluster we do not run the csi rclone driver on all nodes. Just on the ones where user sessions run. But because in this PR amalthea does not have tolerations/affinity the sessions always get scheduled on nodes where the csi rclone driver is not running.

Anyhow thanks for spotting this.

leafty:
> @leafty so I was taking a pointer of a loop variable in the go code. And this will always work how you dont expect it. Because the loop variable is the same for the whole loop duration -not for each loop iteration. So if you have a loop variable and then you take its pointer and store it in a list and use that after the loop has finished what you have is a list of the same variable - i.e. the last value that you looped through.
> 
> This is fixed in go 1.22 but the operator framework is still on 1.21 and I did not want to force an update. There is a super easy fix for this in the code. But it gets me every single time.
> 
> Now you see both pvcs for the datasets: image
> 
> The mounting still fails though because I need #737 so that I can run the sessions on the nodes dedicated for user sessions. On the dev cluster we do not run the csi rclone driver on all nodes. Just on the ones where user sessions run. But because in this PR amalthea does not have tolerations/affinity the sessions always get scheduled on nodes where the csi rclone driver is not running.
> 
> Anyhow thanks for spotting this.
> 
> EDIT: and #737 is the next PR in this stack of PRs.

Since you want to use the address of the value, you could use the index and use the pointer that way, without needing to make copies.

```go
for i := range arr {
  ptr := &arr[i]; // <- no need to copy
}
```

internal/controller/children.go
```diff
@@ -155,7 +155,8 @@ func NewChildResources(cr *amaltheadevv1alpha1.AmaltheaSession) ChildResources {
 
 	desiredDataSourcesPVCs := []ChildResource[v1.PersistentVolumeClaim]{}
 	specPVCs, _, _ := cr.DataSources()
-	for _, desiredPVC := range specPVCs {
+	for i := range specPVCs {
+		desiredPVC := specPVCs[i]
```
leafty, line 159: Note: This makes a copy (which does not seem needed), but is valid code. So no need to change it to use pointers.

olevski: Yeah that makes sense.

leafty (commented):
The fixed loop looks OK :+1:
I can't really comment on the whole PR and I tried to test it, but now I got `storageclass.storage.k8s.io "csi-rclone-387-secret-annotation" not found`.

olevski:
@leafty I purged these old / alternative csi driver installations from dev that I had installed. They were causing trouble in the deployment of a new version of the main rclone csi driver. We cannot run multiple instances in the same cluster. Because when a driver starts it looks at all rclone mounts on every node and it unmounts them. Because when the driver stops the mounts don't work anymore. We will add a functionality to make it so that the mounts are remounted/fixed in the future. But it still is not very easy to see/understand which rclone mounts on the node come from which sci driver installation (when there are many).

olevski:
@sgaist can you re-approve please?

internal/controller/children.go
```diff
@@ -160,21 +175,34 @@ func (c ChildResources) Reconcile(ctx context.Context, clnt client.Client, cr *a
 		Service:     c.Service.Reconcile(ctx, clnt, cr),
 		Ingress:     c.Ingress.Reconcile(ctx, clnt, cr),
 	}
+	dataSrouceUpdates := []ChildResourceUpdate[v1.PersistentVolumeClaim]{}
+	for _, pvc := range c.DataSourcesPVCs {
+		dataSrouceUpdates = append(dataSrouceUpdates, pvc.Reconcile(ctx, clnt, cr))
+	}
+	output.DataSourcesPVCs = dataSrouceUpdates
```
sgaist, line 182:
```suggestion

	dataSourceUpdates := []ChildResourceUpdate[v1.PersistentVolumeClaim]{}
	for _, pvc := range c.DataSourcesPVCs {
		dataSourceUpdates = append(dataSourceUpdates, pvc.Reconcile(ctx, clnt, cr))
	}
	output.DataSourcesPVCs = dataSourceUpdates
```

sgaist (changes requested):
One small typo fix and we will be good

sgaist (approved):
LGTM

PR #723, opened 2024-09-06
feat: Improve visualization of JupyterServer resources for admins.

aledegano:
With this change, when listing JupyterServer, the Renku admin will be able to quickly view the age of the resources which is sometime useful.


aledegano:
I am not sure if something else needs to be touched/regenerated as a result of modifying the CRD, if so please let me know.

olevski (approved):

PR #737, opened 2024-09-23
feat: add tolerations, affinity, node selector

olevski:
This allows the session to be scheduled on a specific nodes. It also allows for tolerations to be specified so that node taints can be handled.


olevski:

- #714
    - #715
        - #737 ◀

api/v1alpha1/amaltheasession_types.go
```diff
@@ -70,6 +70,22 @@ type AmaltheaSessionSpec struct {
...
+
+	// If specified, the pod's scheduling constraints
+	// Passed right through to the Statefulset used for the session.
+	// +optional
+	Affinity *v1.Affinity `json:"affinity,omitempty"`
+
+	// If specified, the pod's tolerations.
+	// Passed right through to the Statefulset used for the session.
+	// +optional
+	Tolerations []v1.Toleration `json:"tolerations,omitempty"`
```
aledegano, line 88: I think I know the answer, but I want to be sure: this is the Kubernetes Toleration "object" and as such Amalthea will support the full definition with `key=value:condition` right?
I am asking because that would be indeed desired to simplify nodes management.

olevski: That is exactly the idea Alessandro. Now that we are in go-land is super easy to pull in well-known components from a Pod spec from the k8s client and just include them in our apispec and have the same validation for them. 

The validation we do may be slightly more permissive in some cases though. Because I dont know if for some of these things k8s has validating/defaulting webhooks that do extra things - and we dont have any mutating or validating webhooks for amalthea. But I think this is not the case for such simple things as tolerations/affinity/node selectors.

olevski: And I plan on adding some defaulting / mutating webhooks at some point in the future. Because without them the default status from the response from the request for creating a session is weird. I.e. you have 0 ready and 0 total containers and in the GET request immediately after you get real numbers.

aledegano: Awesome, thanks for the explanation!

sgaist (approved):

sgaist (approved):

PR #745, opened 2024-10-09
feat: single sidecars exe and versioned templates

olevski:
There are a few things that happen here:
- combine utilities is a single executable that contains the proxy and cloning utilities we run in sidecars
- enable templating of the cloning and auth proxy containers/sidecars that is based on the version of the image passed in the CRD

This later part about "versioned templating" is required to prevent the following scenario:
- we will publish the sidecar image with every release
- the sidecar image is hardcoded as a const in the code
- someone updates this hardcoded constant
- redeploys amalthea on a cluster that already has many sessions
- amalthea restarts and patches all existing images to use the new sidecar image
- this causes all existing running sessions in the cluster to restart

There are other reasons too. For example we upgrade the version of the sidecar image but its api has changed. Then either new sessions with the new sidecar work or the old ones are patched restarted but now when they start back up they keep crashing because the sidecars image API has changed.


internal/controller/amaltheasession_controller.go
```diff
@@ -144,7 +144,18 @@ func (r *AmaltheaSessionReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 
 	log.Info("spec", "cr", amaltheasession)
 
-	children := NewChildResources(amaltheasession)
+	children, err := NewChildResources(amaltheasession)
+	if err != nil {
+		log.Error(
+			err,
+			"Could not template child resources",
```
sgaist, line 151: I am not sure how to interpret that message.

internal/controller/children.go
```diff
@@ -64,6 +64,12 @@ func (c ChildResource[T]) Reconcile(ctx context.Context, clnt client.Client, cr
 				err := ctrl.SetControllerReference(cr, current, clnt.Scheme())
 				return err
 			}
+			for i := range desired.Spec.Template.Spec.InitContainers {
+				cont := &desired.Spec.Template.Spec.InitContainers[i]
+				if cont.Name == "" {
```
sgaist, line 69: Shouldn't that trigger an error ?

sgaist (commented):
It's looking good however I am just not sure about what "template" refers to in this context.

There's the fact that these are template functions but in the k8s context, some resources such as deployment uses template in the sense of "reference pod declaration to be used when creating pods".

internal/controller/children.go
olevski: This is a leftover from development that I should have removed. I will remove it.

internal/controller/amaltheasession_controller.go
olevski: I will revise and clarify it.

olevski:
> It's looking good however I am just not sure about what "template" refers to in this context.

> There's the fact that these are template functions but in the k8s context, some resources such as deployment uses template in the sense of "reference pod declaration to be used when creating pods".

In this case "template" means the real k8s children resources I generate from the CR specification. I.e. the PVC, Ingress, Service, Statefulset, etc.

sgaist (approved):

olevski:

- #745 ◀
    - #748

PR #747, opened 2024-10-18
feat: quieter idleness

sgaist:
## Describe your changes

This PR reduces the number of requests done to the metrics endpoint and lowers the amount of log generated.

The metrics server does not create metrics data until after 10 seconds of pod container lifetime hence requesting them as soon as the pod is created will only show error message for missing pods from the metrics API.

The same goes for the idleness detection, once it was established that a pod is idle, there's no need to further log that it is is idle.

Part of #616

### DIFF
--- a/internal/controller/children.go
+++ b/internal/controller/children.go
@@ -363,7 +363,7 @@ func (c ChildResourceUpdates) Status(
 			if idle && idleSince.IsZero() {
 				idleSince = metav1.NewTime(time.Now())
 			}
-			if !idle && !idleSince.IsZero() {
+			else if !idle && !idleSince.IsZero() {
 				idleSince = metav1.Time{}
 			}
 		}
--- a/internal/controller/children.go
+++ b/internal/controller/children.go
@@ -362,8 +362,7 @@ func (c ChildResourceUpdates) Status(
 			idle = isIdle(ctx, r.MetricsClient, cr)
 			if idle && idleSince.IsZero() {
 				idleSince = metav1.NewTime(time.Now())
-			}
-			else if !idle && !idleSince.IsZero() {
+			} else if !idle && !idleSince.IsZero() {
 				idleSince = metav1.Time{}
 			}
 		}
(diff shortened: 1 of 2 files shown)


internal/controller/children.go
```diff
@@ -271,13 +276,34 @@ func (c ChildResourceUpdates) Conditions(state amaltheadevv1alpha1.State, ctx co
...
+				break
+			}
+		}
+
+		if c.State(cr, pod) == amaltheadevv1alpha1.Running && oldEnough {
+			idle = isIdle(ctx, r.MetricsClient, cr)
+			if idle && idleSince.IsZero() {
+				idleSince = metav1.NewTime(time.Now())
+			}
+			if !idle && !idleSince.IsZero() {
```
Panaetius, line 303: suggestion: 
```suggestion
			else if !idle && !idleSince.IsZero() {
```

internal/controller/status.go
```diff
@@ -118,13 +118,15 @@ func isIdle(ctx context.Context, clnt metricsv1beta1.PodMetricsesGetter, cr *ama
...
 	cpuUsage := metrics.Cpu()
 	if cpuUsage != nil && cpuUsage.Cmp(cpuUsageIdlenessThreshold) == -1 {
-		log.Info(
-			"the session was found to be idle",
-			"cpu usage milicores",
-			cpuUsage.MilliValue(),
-			"idle threshold milicores",
-			cpuUsageIdlenessThreshold.MilliValue(),
-		)
+		if cr.Status.IdleSince.IsZero() {
```
Panaetius, line 121: question: shouldn't this be negated? If I read correctly, in children.go line 304, we set the time to Zero if the session isn't idle. So it this if would only execute when it isn't idle, but then print "session was found to be idle"? or am I misunderstanding the logic here?

Panaetius (changes requested):

internal/controller/status.go
sgaist: Since this check is done at a regular interval, the goal is to have the message printed only when the idleness threshold is reached the first time. Otherwise, each follow up check that would confirm the idleness would print it again.

Panaetius (approved):

PR #748, opened 2024-10-18
feat: add reconcile strategy field in spec

olevski:
feat: add reconcile strategy

chore: cleanup tests


olevski:

- #748 ◀

api/v1alpha1/amaltheasession_types.go
```diff
@@ -65,6 +65,8 @@ type AmaltheaSessionSpec struct {
 
 	// +optional
 	// Additional volumes to include in the statefulset for a session
+	// Volumes used internally by amalthea are all prefixed with 'amalthea-' so as long as you
+	// avoid that naming you will avoid conflicts with the volumes that amalthea generates.
```
sgaist, line 69: Should we have a validator that ensure it does not happen ?

api/v1alpha1/amaltheasession_types.go
```diff
@@ -87,8 +89,19 @@ type AmaltheaSessionSpec struct {
...
+	// +kubebuilder:default:="always"
+	// Indicates how Amalthea should reconcile the child resources for a session. This can be problematic because
+	// newer versions of Amalthea may include new versions of the sidecars or other changes not reflected
+	// in the AmaltheaSession CRD, so simply updating Amalthea could cause existing sessions to restart
+	// because the sidecars will have a newer image or for other reasons because the code changed.
+	// Hibernating the session and deleting it will always work as expected regardless of the strategy.
+	// The status of the session and all hibernation or auto-cleanup functionality will always work as expected.
+	// A few values are possible:
+	// - never: Amalthea will never update any of the child resources and will ignore any changes to the CR
+	// - always: This is the expected method of operation for an operator, changes to the spec are always reconciled
```
sgaist, line 101: Is there a way to notify the user that the session needs to be restarted ? Otherwise we will have the issue of work getting interrupted at the least good moment.

api/v1alpha1/amaltheasession_children.go
```diff
@@ -29,38 +29,14 @@ const sessionVolumeName string = prefix + "volume"
...
-// StatefulSet returns a AmaltheaSession StatefulSet object
-func (cr *AmaltheaSession) StatefulSet() (appsv1.StatefulSet, error) {
-	labels := labelsForAmaltheaSession(cr.Name)
-	replicas := int32(1)
-	if cr.Spec.Hibernated {
-		replicas = 0
-	}
-
-	session := cr.Spec.Session
+func (cr *AmaltheaSession) SessionVolumes() ([]v1.Volume, []v1.VolumeMount) {
```
Panaetius, line 38: suggestion: I am a bit confused why `cr` is the abbreviation for `AmaltheaSession`, especially since in the other file (`auth_templates.go`) it is abbreviated as `session`. I think having a consistent abbreviation can help with code readability.

Panaetius (commented):
just a minor comments. Looks good otherwise, though I agree with Samuel's comments.

api/v1alpha1/amaltheasession_types.go
olevski: @sgaist we will most likely not be running Amalthea with the `always` strategy in renku. I am not sure who you are referring to by "user" because the only "user" we can reliably notify from Amalthea is the admin/service that created the session in k8s. And we can do this via the status field in the CR. But in the case of `always`, notifying the user that the session needs to be restarted makes no sense. The whole point of this strategy is that things will just happen immediately without action from the user or admin.

We can add a notification that the session will be updated with the `whenHibernatedOrFailed` strategy. So that users and admins know that the session is out of date and will be updated when the session is hibernated.

We can do a similar thing for the `never` strategy. 

With always what we can try to do is add something that indicates that the session has been updated. 

I will do this in a follow up PR.

api/v1alpha1/amaltheasession_children.go
olevski: It was meant to stand for "custom resource". I will just rename it to `as` across the board. For "amalthea session".

api/v1alpha1/amaltheasession_types.go
olevski: See #749

api/v1alpha1/amaltheasession_types.go
olevski: See #749 and #751

Panaetius (approved):

Panaetius (approved):

PR #758, opened 2024-11-01
feat: improve error reporting

olevski:
This also adds tests for the most common error cases.

PR #759, opened 2024-11-02
feat: add path prefix for ingress in spec

olevski:


olevski:

- #758
    - #759 ◀

helm-chart/amalthea-sessions/templates/amaltheasession-crd.yaml
```diff
@@ -4011,6 +4011,12 @@ spec:
                       rule: self == oldSelf
                   ingressClassName:
                     type: string
+                  pathPrefix:
```
m-alisafaee, line 4014: Shouldn't these modifications be done in `bundle/manifests/amalthea.dev_amaltheasessions.yaml` as well?

test/utils/utils.go
```diff
@@ -58,7 +58,7 @@ const (
 )
 
 func warnError(err error) {
-	fmt.Fprintf(GinkgoWriter, "warning: %v\n", err)
+	GinkgoLogr.Error(err, "warning error occurred")
```
m-alisafaee, line 61: Should we remove `GinkgoWriter` initialization from `internal/controller/suite_test.go:57`?

api/v1alpha1/amaltheasession_types.go
```diff
@@ -161,6 +162,11 @@ type Ingress struct {
 	// +optional
 	// The name of the TLS secret, same as what is specified in a regular Kubernetes Ingress.
 	TLSSecret *SessionSecretRef `json:"tlsSecret,omitempty"`
+	// +optional
+	// The path prefix that will be used in the ingress. If left unset or set to "" then
```
m-alisafaee, line 166: What is the default value when it isn't set? I'm not sure if it can be `nil` since it's not a pointer. If it can be `nil` then you should check against it as well in the above file code: `if cr.Spec.Ingress.PathPrefix == "" {`; if not, then please change this comment to say what's the default value.

m-alisafaee (commented):
Thanks Tasko! Please see my comments.

api/v1alpha1/amaltheasession_children.go
```diff
@@ -201,6 +201,17 @@ func (cr *AmaltheaSession) sessionLocalhostURL() *url.URL {
 	return &output
 }
 
+// The localhost path prefix URL for the session. It excludes the auth proxy and the ingress and
+// the host is always 127.0.0.1.
+func (cr *AmaltheaSession) localhostPathPrefixURL() *url.URL {
+	if cr.Spec.Ingress.PathPrefix == "" {
+		return cr.sessionLocalhostURL()
+	}
```
leafty, line 209: This branch sounds like it would make it very easy to not understand how to properly configure an Amalthea Session.

If the default URL for my session is `/lab`, then the ingress will fully restrict the user to `/lab` URLs, but actually, some other paths may be valid, e.g `/tree` and `/notebooks` are valid paths when jupyter server is running.

api/v1alpha1/amaltheasession_types.go
olevski: The default for a string when it is not set is `""` - just empty string. The default for a pointer that is not set is `nil`.

api/v1alpha1/amaltheasession_children.go
olevski: Yeah I can make it `/`.

test/utils/utils.go
olevski: No because when you set the logging to very verbose then it prints the output from the operator. And that line allows it to show up. So we can leave that.

api/v1alpha1/amaltheasession_types.go
olevski: But as @leafty suggested we can just set the default to `/` that works better.

leafty (approved):

leafty (approved):

PR #760, opened 2024-11-04
fix: authproxy CLI args parsing

olevski:
The parsing of the CLI args and the inclusion of parameters from environment variables and config file was a bit off. Not it works as expected.


olevski:

- #758
    - #759
        - #760 ◀

internal/authproxy/serve.go
```diff
@@ -156,6 +154,23 @@ func serve(cmd *cobra.Command, args []string) {
 	}
 	e.Use(middleware.Proxy(middleware.NewRoundRobinBalancer(targets)))
 
-	e.Logger.Info(fmt.Sprintf("Starting proxy for %v", remoteURL))
+	e.Logger.Infof("Starting proxy for remote: %s, cookie key: %s, token of length %d", remoteURL.String(), cookieKey, len(token))
```
sgaist, line 157: Shouldn't the port also be part of the printed information ?

olevski: No because when the server starts it prints the port. There is no need to print it twice.

internal/authproxy/serve.go
```diff
@@ -32,76 +37,90 @@ import (
...
 
-	serveCmd.PersistentFlags().Int(PortFlag, 65535, "port on which the proxy will listen")
-	err = viper.BindPFlag(PortFlag, serveCmd.PersistentFlags().Lookup(PortFlag))
+	serveCmd.PersistentFlags().IntVar(&port, portFlag, 65535, "port on which the proxy will listen")
+	err = viper.BindPFlag(prefix+"."+portFlag, serveCmd.PersistentFlags().Lookup(portFlag))
 	if err != nil {
 		return nil, err
 	}
-	err = viper.BindEnv(PortFlag)
+	err = viper.BindEnv(prefix+"."+portFlag, strings.ToUpper(prefix+"_"+portFlag))
```
sgaist, line 83: I just realized, aren't your manually doing what `Viper.SetEnvPrefix("authproxy")` and `viper.AutomaticEnv()` do ?

See [here](https://github.com/spf13/viper?tab=readme-ov-file#env-example) for what I had in mind.

olevski: This was such a pain to debug. And for some reason I could not get it to work with flags, the files and env variables when `viper.SetEnvPrefix` was used. 

The reason for this is that when you use AutomaicEnv and SetEnvPrefix then it assumes that the name of the variable in viper equals the name of the env variable. But this is not the case. We want the names in viper to have a prefix like `section.name` so that they can be loaded from a yaml file like 

```yaml
section:
  name: secret-value
```

But when you use automatic env then it puts the name of the variable in viper under just `name`. It removes the prefix. So then flags and the secret file get saved in a different name in viper compared to env secrets. This was the only way I could work out to make sure that we have sections in the yaml file for different things - not just one flat dictionary and to have the secrets and the flags all work and populate the same place in viper.

olevski: In the future I will remove the options to configure the sidecar container through so many things. We dont have to accept cli args, config file and env vars all at the same time and try to merge all of them.

sgaist: Ok. all good then, thanks.

sgaist (dismissed):

olevski:
@sgaist this is ready now. I noticed that the exit codes of the tests failing were not properly propagating to the github actions. So I fixed this now.

sgaist (approved):

PR #778, opened 2024-12-04
fix: clear status after hibernating or resuming

olevski:
Some status fields - most notably, the timestamps that indicate how long the session has been failing, hibernated or idle should be cleared after hibernation or resuming. If this is not done then one cannot successfully resume the session after hibernation (for example).

### DIFF
--- a/internal/controller/amaltheasession_controller_test.go
+++ b/internal/controller/amaltheasession_controller_test.go
@@ -431,7 +431,7 @@ var _ = Describe("AmaltheaSession Controller", func() {
 			Expect(err).NotTo(HaveOccurred())
 			Expect(k8sClient.Get(ctx, typeNamespacedName, amaltheasession)).To(Succeed())
 
-			By("Aritificially adding statuses")
+			By("Artificially adding statuses")
 			newTimestamp := metav1.NewTime(time.Now().Round(time.Second))
 			amaltheasession.Status.FailingSince = newTimestamp
 			amaltheasession.Status.IdleSince = newTimestamp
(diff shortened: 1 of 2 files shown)


internal/controller/amaltheasession_controller_test.go
```diff
@@ -421,5 +421,83 @@ var _ = Describe("AmaltheaSession Controller", func() {
...
+		It("should clear statuses after hibernation", func(ctx SpecContext) {
+			By("Checking if the custom resource was successfully created")
+			controllerReconciler := newReconciler()
+			_, err := controllerReconciler.Reconcile(ctx, reconcile.Request{
+				NamespacedName: typeNamespacedName,
+			})
+			Expect(err).NotTo(HaveOccurred())
+			Expect(k8sClient.Get(ctx, typeNamespacedName, amaltheasession)).To(Succeed())
+
+			By("Aritificially adding statuses")
```
Panaetius, line 434: spelling

internal/controller/amaltheasession_controller_test.go
```diff
@@ -421,5 +421,83 @@ var _ = Describe("AmaltheaSession Controller", func() {
...
+			By("Marking the session as hibernated")
+			amaltheasession.Spec.Hibernated = true
+			Expect(controllerReconciler.Update(ctx, amaltheasession)).To(Succeed())
+			_, err = controllerReconciler.Reconcile(ctx, reconcile.Request{
+				NamespacedName: typeNamespacedName,
+			})
+			Expect(err).NotTo(HaveOccurred())
+			Expect(k8sClient.Get(ctx, typeNamespacedName, amaltheasession)).To(Succeed())
+
+			By("Aritificially adding statuses")
```
Panaetius, line 477: spelling

internal/controller/children.go
```diff
@@ -368,6 +386,29 @@ func Conditions(
 	return conditions
 }
 
+// statusCallback modifies an amalthea status based on the reconciliation of different child resources.
+// Certain status updates can only be done based on the current and previous values of child resources,
+// these should take precedence over other statuses dervied only on current values and are applied here.
```
Panaetius, line 391:
```suggestion
// these should take precedence over other statuses derived only on current values and are applied here.
```

Panaetius (changes requested):
only some spelling mistakes, otherwise looks great

Panaetius (approved):

PR #796, opened 2024-12-13
fix: typo in jupyter server service template

olevski:
I suspect this was just ignored by k8s up until  now. With this set the network request goes directly to the session pod skipping some proxying that would happen if we did not have this set to None. But now I noticed that k8s fails.


olevski:
See https://kubernetes.io/docs/concepts/services-networking/service/#headless-services

aledegano (dismissed):

aledegano (approved):

PR #854, opened 2025-01-29
Add csi-rclone as a direct subchart

sambuc:


helm-chart/amalthea-sessions/values.yaml
```diff
@@ -36,3 +36,16 @@ sidecars:
   image:
     repository: renku/sidecars
     tag: latest
+csi-rclone:
+  install: false
+  # This section is only relevant if you are installing csi-rclone as part of Amalthea.
+  # When this chart is deployed as part of Renkulab the following value should match the name of the storage class
+  # configured for Notebooks cloudstorage.
+  ## Name of the csi storage class to use for RClone/Cloudstorage. Should be unique per cluster.
+  # storageClassName: csi-rclone
```
sgaist, line 45: I think that one should be `csi-rclone-secret-annotation` since it's the method used to for the defining the configuration.

olevski: I think we should remove all the commented stuff. Just point people to https://github.com/SwissDataScienceCenter/csi-rclone/blob/master/deploy/csi-rclone/values.yaml so that they check what values are available. And be default the default values of csi-rclone should work.

helm-chart/amalthea-sessions/Chart.yaml
```diff
@@ -19,3 +19,9 @@ version: 0.1.0
 # follow Semantic Versioning. They should reflect the version the application is using.
 # It is recommended to use it with quotes.
 appVersion: "0.1.0"
+dependencies:
+  - name: csi-rclone
+    repository: "https://swissdatasciencecenter.github.io/helm-charts/"
+    version: "0.3.5"
+    # csi-rclone.install instruct to also deploy csi-rclone with amalthea-sessions.
+    condition: csi-rclone.install
```
olevski, line 27:
```suggestion
    condition: dependencies.csi-rclone.install
```

helm-chart/amalthea-sessions/values.yaml
```diff
@@ -36,3 +36,16 @@ sidecars:
...
+  # csiNodepluginRclone:
+  #   nodeSelector: {}
+  # Set tolerations if you have taints on your user session nodes. The csi has to run on every node
+  # where it is used.
+  #   tolerations: []
+  #   affinity: {}
```
olevski, line 51:
```suggestion
dependencies:
  csi-rclone: false
```

olevski (changes requested):

helm-chart/amalthea-sessions/values.yaml
sambuc: I am not using the `dependencies` key as suggested because this is the array of child helm charts repositories.

helm-chart/amalthea-sessions/values.yaml
sambuc: I removed the comments and updated the default value to match the current default value of the child helm chart.

helm-chart/amalthea-sessions/templates/_helpers.tpl
```diff
@@ -1,3 +1,11 @@
+{{/*
+Check consistency requirements.
+*/}}
+{{- $expectedName := printf "%s-%s" .Values.csiRclone.storageClassName "secret-annotation" -}}
+{{- if not eq .Values.rcloneStorageClass $expectedName -}}
+{{- fail "ERROR: .Values.rcloneStorageClass does not match " $expectedName ". Please Refer to csi-rclone documentation." }}
+{{- end }}
```
olevski, line 7: It makes sense to run this check also only when we deploy the csi rclone chart as part of amalthea. If not then we should not check. So I just add an outer if statement for this. You can use `and` but I dont remember the right syntax and you will need some brackets when you combine that with `not eq` in the same if statement. So I think the suggestion below is good enough and potentially clearer than the combined if statement.

```suggestion
{{- if .Values.deploy.csi-rclone }}
{{- if not eq .Values.rcloneStorageClass $expectedName }}
{{- fail "ERROR: .Values.rcloneStorageClass does not match " $expectedName ". Please Refer to csi-rclone documentation." }}
{{- end }}
{{- end }}
```

olevski (changes requested):

olevski (approved):

olevski:
The failing test is not your fault. I think we have some flakiness there. I reran it. Hopefully it passes.

olevski:
@sambuc I was wrong before. The test that is failing is failing because the utility function that installs the helm chart for the tests now needs to run `helm build dep` or something like that. This is mentioned in the error message. Before we did not have any dependencies so it did not matter. But now we do so we need that command.

sambuc:
My apologies @olevski, as we were talking about on slack, I missed the fact you had already patched it here as well.

olevski (approved):
:rocket: Thanks Lionel!

helm-chart/amalthea-sessions/values.yaml
sambuc: Thank you for spotting this and your comment @sgaist.

There was more to it as you might have seen from the several corrections I had to apply on top of my first patch.

Issue #751, opened 2024-10-21
Surface errors from child resources in the CR status.


olevski:
Right now when a statefulset for amalthea is invalid or fails to start we do not surface this error in the amaltheaSession status.

But we should. Especially for cases that are not transient and require intervention from the user/creator of the session to fix.

olevski:
Followup from #748

PR #858, opened 2025-01-30
refactor: improve error reporting with regard to image pulling

sgaist:
## Describe your changes

Information related to image pulling errors were lost in the original implementation because not all related cases related were handled the same.

Now the original error message is passed back in all cases.

Part of #751


olevski (approved):
That is a nice change. It makes the code much more readable. Thank you Samuel!

PR #861, opened 2025-01-31
feat: improve unschedulable error report

sgaist:
## Describe your changes

This PR refactors the handling of unschedulable pods and propagate the information back to the AmaltheaSession.

There are multiple possible reasons for pods not to be scheduled:

- Resource requests outside of system capabilties
- Topological reason (i.e. no nodes matching the selector desired)

As a drive by cleanup, the handling of the helm chart and external resources has also been improved to:
- Allow multiple runs
- Ignore not found resources when deleting

### DIFF
--- a/internal/controller/children.go
+++ b/internal/controller/children.go
@@ -328,7 +328,7 @@ func (c ChildResourceUpdates) failureMessage(pod *v1.Pod) string {
 func (c ChildResourceUpdates) warningMessage(pod *v1.Pod) string {
 	for _, condition := range pod.Status.Conditions {
 		if condition.Reason == "Unschedulable" {
-			return fmt.Sprintf("the session cannot be scheduled due to: %s", condition.Message)
+			return fmt.Sprintf("the session cannot be scheduled due to: %s. Please contact an administrator.", condition.Message)
 		}
 	}
 	return ""
--- a/internal/controller/children.go
+++ b/internal/controller/children.go
@@ -328,7 +328,7 @@ func (c ChildResourceUpdates) failureMessage(pod *v1.Pod) string {
 func (c ChildResourceUpdates) warningMessage(pod *v1.Pod) string {
 	for _, condition := range pod.Status.Conditions {
 		if condition.Reason == "Unschedulable" {
-			return fmt.Sprintf("the session is requesting more resources than available, the reason is %q, this may correct itself if a Renku admin provisions more resources", condition.Reason)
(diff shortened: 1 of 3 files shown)


internal/controller/children.go
```diff
@@ -328,7 +328,7 @@ func (c ChildResourceUpdates) failureMessage(pod *v1.Pod) string {
 func (c ChildResourceUpdates) warningMessage(pod *v1.Pod) string {
 	for _, condition := range pod.Status.Conditions {
 		if condition.Reason == "Unschedulable" {
-			return fmt.Sprintf("the session is requesting more resources than available, the reason is %q, this may correct itself if a Renku admin provisions more resources", condition.Reason)
```
olevski, line 331: question: why change this message? It is true that the problem may correct itself if more resources are provisioned. I think we should remove the `Renku admin` part and just say `administrator` or `IT administrator`.

olevski (changes requested):

internal/controller/children.go
sgaist: Because it might be misleading. There are multiple possible cause for pods being unschedulable:
- Resource requests that cannot be fulfilled as hints the original error message
- Topology constraints such that the pod cannot be deployed with the specified selector
This message only takes into account the first case.
That said, I agree, both cases require administrator intervention.
I'll update it.

aledegano: If I may add: the chance of an admin intervention is pretty unlikely and will probably give the wrong message.

The probability of an admin being able to throw even more resources in the pool is very low.

The most likely event that might unlock a session that cannot schedule because of lack of resources is actually another session stopping.

Because of the above I would remove the "admin" entirely from the conversation...

sgaist: @aledegano while I agree on the conclusion (more hardware getting added is unlikely). This message will be seen by the user so suggesting to contact an administrator but without specifying what the solution might still be better experience than just: `your session wont start: difficult to decipher k8s error`.

olevski (approved):

internal/controller/children.go
olevski: > If I may add: the chance of an admin intervention is pretty unlikely 

We dont know this. In our case adding resources is very hard and unlikely to happen. But in other cases it may not be. I think in this case the possible resolutions are:
- wait for more space to free up (this is not always guaranteed because the cluster may be misconfigured or missing resources fully so that no matter how long you wait nothing will change).
- restart the session with fewer resource requests 
- contact the admin to add more resources

So I think it makes sense to include this in the message.

PR #895, opened 2025-03-05
feat: surface quota exceede errors on crd

Panaetius:
trying to "Owns()" events doesn't work, you don't get stateful set events..

The current solution works, though might have performance implications due to fetching events for each statefulset in the reconcile loop.

The status on a session would look like:
```
status:
  conditions:
    - lastTransitionTime: '2025-03-07T09:35:12Z'
      message: The session is hibernated
      reason: Hibernated
      status: 'False'
      type: Ready
    - lastTransitionTime: '2025-03-06T09:30:28Z'
      message: The ingress is setup and operational
      reason: IngressOperational
      status: 'True'
      type: RoutingReady
  containerCounts: {}
  error: Quota exceeded
  idle: false
  initContainerCounts: {}
  state: Failed
  url: https://renku-ci-am-895.dev.renku.ch/sessions/john-doe-01adadc97abf/lab/
```

### DIFF
--- a/internal/controller/children.go
+++ b/internal/controller/children.go
@@ -12,7 +12,6 @@ import (
 	networkingv1 "k8s.io/api/networking/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -101,26 +100,6 @@ func (c ChildResource[T]) Reconcile(ctx context.Context, clnt client.Client, cr 
 			desired, ok := any(c.Desired).(*appsv1.StatefulSet)
 			if !ok {
 				return fmt.Errorf("Could not cast when reconciling")
-			}
-			events := v1.EventList{}
-			selector, err := fields.ParseSelector(fmt.Sprintf("involvedObject.name=%s,involvedObject.namespace=%s,involvedObject.kind=StatefulSet", current.ObjectMeta.Name, current.ObjectMeta.Namespace))
-			if err == nil {
-				if err := clnt.List(ctx, &events, &client.ListOptions{FieldSelector: selector}); err == nil {
-					for _, event := range events.Items {
-						if event.Reason == "FailedCreate" && strings.Contains(event.Message, "exceeded quota") {
-							statusCallback = func(status *amaltheadevv1alpha1.AmaltheaSessionStatus) {
-								status.State = amaltheadevv1alpha1.Failed
-								status.Error = "Quota exceeded"
-							}
-							return nil
-						}
-					}
-				} else {
-					log.Error(err, "couldn't list events")
-
-				}
-			} else {
-				log.Error(err, "couldn't parse selector")
 			}
 			if current.CreationTimestamp.IsZero() {
 				log.Info("Creating a statefulset")
@@ -437,7 +416,7 @@ func (c ChildResourceUpdates) Status(
 ) amaltheadevv1alpha1.AmaltheaSessionStatus {
 	log := log.FromContext(ctx)
 
-	pod, err := cr.Pod(ctx, r.Client)
+	pod, err := cr.GetPod(ctx, r.Client)
 	if err != nil && !apierrors.IsNotFound(err) {
 		log.Error(err, "Could not read the session pod when updating the status")
 	}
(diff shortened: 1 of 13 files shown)


olevski:
@Panaetius I am writing this down as per our discussion and just for future reference.

A big part of the guts of the controller is `manager.Manager` from the `controllerruntime` k8s package. We initialize this via [NewManager](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.20.3/pkg/manager#New). In there there is a `Cache` field in the [Options](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.20.3/pkg/manager#Options). This `Cache` field is itself another set of [CacheOptions](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.20.3/pkg/cache#Options).

We specify the cache Options as follows:

```
	cacheOptions := cache.Options{
		DefaultNamespaces: map[string]cache.Config{releaseNamespace: {}},
	}
	if clusterScoped {
		cacheOptions = cache.Options{
			DefaultNamespaces: nil,
		}
	}
```

Which means we leave the `ReaderFailOnMissingInformer` field of the Cache options as the default:

```
	// ReaderFailOnMissingInformer configures the cache to return a ErrResourceNotCached error when a user
	// requests, using Get() and List(), a resource the cache does not already have an informer for.
	//
	// This error is distinct from an errors.NotFound.
	//
	// Defaults to false, which means that the cache will start a new informer
	// for every new requested resource.
	ReaderFailOnMissingInformer bool
```

So that means that if we did not already have an Informer (i.e. this is essentially the thing that watches events and keeps an in-memory cache) then one would be created. So if we were to roll this out we should not overload the API.

internal/controller/children.go
```diff
@@ -101,14 +102,34 @@ func (c ChildResource[T]) Reconcile(ctx context.Context, clnt client.Client, cr
...
+							return nil
+						}
+					}
+				} else {
+					log.Error(err, "couldn't list events")
+
+				}
+			} else {
+				log.Error(err, "couldn't parse selector")
+			}
```
olevski, line 124: suggestion: You can use `client.MatchingFields{}` which can be constructed like a Set and it cannot error out. Rather than parse the selector and then check for error.

See the end of the example in: https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client#FieldIndexer

internal/controller/children.go
```diff
@@ -101,14 +102,34 @@ func (c ChildResource[T]) Reconcile(ctx context.Context, clnt client.Client, cr
 			if !ok {
 				return fmt.Errorf("Could not cast when reconciling")
 			}
+			events := v1.EventList{}
```
olevski, line 105: issue: This whole change should be moved to the `State` method on `ChildResourceUpdates` rather than here. The statusCallback function is meant only for cases where the status depends on the previous and current state - like hibernation. You do not need to know the previous state for this case so you do not need to do this here.

You can get the k8s client from `r *AmaltheaSessionReconciler` in `Status()` and then feed it through to `State()` which is called from within `Status()`.

internal/controller/children.go
```diff
@@ -101,14 +102,34 @@ func (c ChildResource[T]) Reconcile(ctx context.Context, clnt client.Client, cr
...
+			selector, err := fields.ParseSelector(fmt.Sprintf("involvedObject.name=%s,involvedObject.namespace=%s,involvedObject.kind=StatefulSet", current.ObjectMeta.Name, current.ObjectMeta.Namespace))
+			if err == nil {
+				if err := clnt.List(ctx, &events, &client.ListOptions{FieldSelector: selector}); err == nil {
+					for _, event := range events.Items {
+						if event.Reason == "FailedCreate" && strings.Contains(event.Message, "exceeded quota") {
+							statusCallback = func(status *amaltheadevv1alpha1.AmaltheaSessionStatus) {
+								status.State = amaltheadevv1alpha1.Failed
+								status.Error = "Quota exceeded"
```
olevski, line 113: suggestion: Please add a more descriptive message here that a not-so-technical user would understand. If I saw this message I would be wondering what the quota refers to (i.e. for compute resources). Also what can I do to fix this. We can provide some general guidance here. For example request less resources or contact your administrator.

internal/controller/children.go
```diff
@@ -101,14 +102,34 @@ func (c ChildResource[T]) Reconcile(ctx context.Context, clnt client.Client, cr
 			if !ok {
 				return fmt.Errorf("Could not cast when reconciling")
 			}
+			events := v1.EventList{}
+			selector, err := fields.ParseSelector(fmt.Sprintf("involvedObject.name=%s,involvedObject.namespace=%s,involvedObject.kind=StatefulSet", current.ObjectMeta.Name, current.ObjectMeta.Namespace))
+			if err == nil {
```
aledegano, line 107: If you reverse some check you could have much less nested code, e.g.:
```
if err != nil {
  log.Error(...)
}
if err := clnt.List[....]
```

internal/controller/children.go
```diff
@@ -101,14 +102,34 @@ func (c ChildResource[T]) Reconcile(ctx context.Context, clnt client.Client, cr
...
+				if err := clnt.List(ctx, &events, &client.ListOptions{FieldSelector: selector}); err == nil {
+					for _, event := range events.Items {
+						if event.Reason == "FailedCreate" && strings.Contains(event.Message, "exceeded quota") {
+							statusCallback = func(status *amaltheadevv1alpha1.AmaltheaSessionStatus) {
+								status.State = amaltheadevv1alpha1.Failed
```
olevski, line 112: suggestion: please add a test for finding the quota exceeded failure in the `When the session is failing` context in `test/e2e/general_test.go`. You can see examples for other tests for failing sessions there and follow the same pattern for this.

olevski (changes requested):
Thanks for addressing this Ralf. I added a few comments. 

Also please run `make lint` and `make lint-fix`. I think this should clear some of the failing tests.

internal/controller/children.go
Panaetius: I on purpose have else statements because I don't error out on errors but want to continue normally. If for some reason we can't get events, the other code still works. I mean, quota errors happen in <5% of the cases and we wouldn't want to have the operator refuse to concilliate and break if it can't get events for whatever reason. It'd still do its thing in the >95% other cases. hence just logging an error and carrying on.

But maybe I'm missing something?

aledegano: my bad, I thought that `log.error` would also return.
So, unfortunately I don't have any practical suggestion, the only feedback is that it's quite hard to parse so many nested (and thus indented) cases. If you see a way to streamline it would be great, otherwise never mind :)

olevski (approved):

PR #905, opened 2025-03-20
feat: add image pull policy to the CRD

olevski:
Adds a new field to the CRD.

The default is `IfNotPresent` because  that is what we had hardcoded before in the code - so this way we remain compatible.


sgaist (dismissed):

leafty (commented):
Note that `IfNotPresent` as a default is not safe!

olevski:
@leafty I switched the default to `Always`.

sgaist (approved):

PR #931, opened 2025-05-14
chore: update to go 1.24

olevski:
This will allow us the rest of the updates from dependabot to merge. There is some weird problem with golangci lint because it is using 1.24 but  the project isnt or vice-versa.

Plus it is good to be up to date.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,8 +1,6 @@
 module github.com/SwissDataScienceCenter/amalthea
 
-go 1.23.0
-
-toolchain go1.23.2
+go 1.24.3
 
 require (
 	github.com/go-git/go-git/v5 v5.14.0
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -22,7 +22,7 @@
 		"ghcr.io/devcontainers-contrib/features/poetry:2": {},
 		"ghcr.io/dhoeric/features/k9s:1": {},
 		"ghcr.io/devcontainers/features/go:1": {
-			"version": "1.23"
+			"version": "1.24"
 		}
 	},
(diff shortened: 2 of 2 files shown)


eikek (approved):

go.mod
```diff
@@ -1,8 +1,6 @@
 module github.com/SwissDataScienceCenter/amalthea
 
-go 1.23.0
-
-toolchain go1.23.2
+go 1.24.3
```
eikek, line 3: I'm just seeing the `toolchain` is removed - I assume on purpose, just pointing it out in case it was an oversight :-) From what I was reading one could use `go 1.24.0` and `toolchain 1.24.3`. However, I have no practical knowledge :-)

olevski: No I have no clue what that was for. I think it is an old thing not used anymore. I did not even remove it. I ran `go mod tidy` after updating the go version which is supposed to cleanup unused dependencies. And it removed that line automatically.

olevski: From the docs:

> The toolchain line declares a suggested toolchain to use with the module or workspace. As described in “[Go toolchain selection](https://go.dev/doc/toolchain#select)” below, the go command may run this specific toolchain when operating in that module or workspace if the default toolchain’s version is less than the suggested toolchain’s version. If the toolchain line is omitted, the module or workspace is considered to have an implicit toolchain goV line, where V is the Go version from the go line.

So it falls back to the go version line in the go.mod. And since for us they were they same it is not needed.

eikek: Thank you. I was reading something about the `go …` line meaning "I'm using language features of that go version" and `toolchain …` to be "use this for building". Though, I was confused by this. I was thinking why you would want to separate this, it could lead to weird misconfigurations, but anyways, I didn't know so was mentioning it here.

PR #938, opened 2025-05-21
feat: track last request time in auth proxy for idle check

Panaetius:
/deploy renku=amalthea-idleness-requests extra-values=notebooks.culling.idleSecondsThreshold.registered=300

I think this would break existing sessions on roll out, as amalthea doesn't have permission to `patch` resources of type `service`

### DIFF
--- a/internal/authproxy/serve.go
+++ b/internal/authproxy/serve.go
@@ -144,17 +144,19 @@ func Command() (*cobra.Command, error) {
 	return serveCmd, nil
 }
 
-type (
-	RequestStats struct {
-		LastRequest time.Time
-		mutex       sync.RWMutex
-	}
-)
+type RequestStats struct {
+	LastRequest time.Time
+	mutex       sync.RWMutex
+}
 
 func NewStats() *RequestStats {
 	return &RequestStats{
 		LastRequest: time.Now(),
 	}
+}
+
+type RequestStatsResponse struct {
+	LastRequestTime time.Time `json:"last_request_time"`
 }
 
 func (l *RequestStats) Process(next echo.HandlerFunc) echo.HandlerFunc {
@@ -172,7 +174,7 @@ func (l *RequestStats) Process(next echo.HandlerFunc) echo.HandlerFunc {
 func (l *RequestStats) Handle(c echo.Context) error {
 	l.mutex.RLock()
 	defer l.mutex.RUnlock()
-	return c.String(http.StatusOK, l.LastRequest.Format("2006-01-02 15:04:05"))
+	return c.JSON(http.StatusOK, RequestStatsResponse{LastRequestTime: l.LastRequest})
 }
 
 func serve(cmd *cobra.Command, args []string) {
--- a/internal/controller/children.go
+++ b/internal/controller/children.go
@@ -3,8 +3,6 @@ package controller
 import (
 	"context"
 	"fmt"
-	"io"
-	"net/http"
 	"strings"
 	"time"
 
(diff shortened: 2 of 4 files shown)


internal/authproxy/serve.go
```diff
@@ -132,6 +144,37 @@ func Command() (*cobra.Command, error) {
 	return serveCmd, nil
 }
 
+type (
+	RequestStats struct {
+		LastRequest time.Time
+		mutex       sync.RWMutex
+	}
+)
```
olevski, line 152: Nitpick: What you have is fine, it is just unnecessary.

```suggestion
type RequestStats struct {
	LastRequest time.Time
	mutex       sync.RWMutex
}

```

internal/authproxy/serve.go
```diff
@@ -132,6 +144,37 @@ func Command() (*cobra.Command, error) {
...
+	return &RequestStats{
+		LastRequest: time.Now(),
+	}
+}
+
+func (l *RequestStats) Process(next echo.HandlerFunc) echo.HandlerFunc {
+	return func(c echo.Context) error {
+
+		if err := next(c); err != nil {
+			c.Error(err)
```
olevski, line 164: question: I think you can simply return the error, right? I dont see why you have to wrap it like that. I think the central error handler in echo will handle things properly if you just return the error.

```suggestion
			return err
```

internal/authproxy/serve.go
```diff
@@ -132,6 +144,37 @@ func Command() (*cobra.Command, error) {
...
+		l.mutex.Lock()
+		defer l.mutex.Unlock()
+		l.LastRequest = time.Now()
+		return nil
+	}
+}
+func (l *RequestStats) Handle(c echo.Context) error {
+	l.mutex.RLock()
+	defer l.mutex.RUnlock()
+	return c.String(http.StatusOK, l.LastRequest.Format("2006-01-02 15:04:05"))
```
olevski, line 175: suggestion: I know we briefly talked about this and I originally agreed to this. But after thinking about it, it is better to make this a json dict. That way we can update this in a backwards compatible way by adding different fields and keeping the old fields around. Whereas with just a string like this pretty much any change we make will be a breaking one. And we want this to be robust because a new version of amalthea will be calling the endpoints on old sessions that are just hanging around. For now it will have 1 single key but that is ok. You can make it a `map[string]string` and serialize that - you dont need a whole new struct just for this.

olevski: Also lets use a stndard for the datetime format. In go you have builtin `time.RFC3339` which you can pass straight to `.Format()`

internal/controller/children.go
```diff
@@ -459,6 +461,61 @@ func (c ChildResourceUpdates) statusCallback(status *amaltheadevv1alpha1.Amalthe
...
+	if err != nil {
+		return time.Time{}, fmt.Errorf("Couldn't parse date from url %s : %w", url, err)
+	}
+	return rt, nil
+}
+
+func getIdleSince(
+	ctx context.Context,
+	r *AmaltheaSessionReconciler,
+	cr *amaltheadevv1alpha1.AmaltheaSession,
```
olevski, line 494: suggestion: I think we should just merge `isIdle` and `getIdleSince` into one function. They only depend on one another and `getIdleSince` also returns whether the function is idle not just the time. So it can be confusing for developers who are looking at these 2 functions which one to call. And I think the logic is all tied together anyhow.

internal/controller/children.go
```diff
@@ -459,6 +461,61 @@ func (c ChildResourceUpdates) statusCallback(status *amaltheadevv1alpha1.Amalthe
...
+		return time.Time{}, fmt.Errorf("Couldn't get last request time: %s", body)
+	}
+	rt, err := time.Parse("2006-01-02 15:04:05", string(body))
+	if err != nil {
+		return time.Time{}, fmt.Errorf("Couldn't parse date from url %s : %w", url, err)
+	}
+	return rt, nil
+}
+
+func getIdleSince(
```
sgaist, line 491: Since the method returns two different, yet related, information (idle since and idle), I think `getIdleSince` is a bit misleading now.
Maybe something like:
```suggestion
func getIdleState(
```

olevski (changes requested):
So some of my comments overlap with Samuel. But I left them in. Thanks Ralf this looks good!

internal/authproxy/serve.go
Panaetius: I don't really know, I was following the [middleware cookbook](https://echo.labstack.com/docs/cookbook/middleware)

sgaist: `Error` calls the HttpErrorHandler method so I think it's mainly about handling `echo` related errors that are not related to the error that might be returned by the rest of the code.

olevski (dismissed):

olevski (approved):

sgaist (approved):

PR #939, opened 2025-05-22
chore: revert python updates to kopf and kubernetes

olevski:
/deploy


olevski:
I think the main culprits for the problems are the kubernetes and/or kopf library. So those were definitely were reverted.

PR #952, opened 2025-06-18
refactor:crd deployment

sgaist:
## Describe your changes

This PR refactors the CRD deployment part of the amalthea and amalthea-session helm charts.

The management of the CRD, while looking simple, is in fact quite complex in its consequences.
Hence we will follow the current recommendation of the Helm maintainers.
More information about the reasoning can be found at:

https://github.com/helm/community/blob/main/hips/hip-0011.md

The RBAC cleanup will also allow for deployment on OpenShift.

### DIFF
--- a/helm-chart/amalthea/templates/rbac/_rbac_rules.tpl
+++ b/helm-chart/amalthea/templates/rbac/_rbac_rules.tpl
@@ -26,15 +26,21 @@
   # Note that we do not patch/update/delete them ever.
   - apiGroups:
       - ""
-      - apps
-      - networking.k8s.io
     resources:
-      - statefulsets
       - persistentvolumeclaims
       - services
-      - ingresses
       - secrets
       - configmaps
+    verbs: [create, get, list, watch]
+  - apiGroups:
+      - apps
+    resources:
+      - statefulsets
+    verbs: [create, get, list, watch]
+  - apiGroups:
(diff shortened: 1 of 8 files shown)


helm-chart/amalthea/templates/rbac/_rbac_rules.tpl
```diff
@@ -26,16 +26,22 @@
   # Note that we do not patch/update/delete them ever.
   - apiGroups:
       - ""
-      - apps
-      - networking.k8s.io
     resources:
-      - statefulsets
       - persistentvolumeclaims
       - services
-      - ingresses
```
olevski, line 35: question: The removal of these and adding them as separate entries just means that openshift does not like them combined like this when you have more than 1 group?

olevski (changes requested):
Just have one question about the RBAC changes.

helm-chart/amalthea/templates/rbac/_rbac_rules.tpl
sgaist: The combination ends up with trying to create rules that are invalid in the sense that there's no statefulsets in the default API nor in `networking.k8s.io` so you will get an error (at least as normal user) telling you that you are trying to create rules for which you don't have authority to.

PR #955, opened 2025-06-19
fix: remove extra auth container

### LINKED ISSUE #954: bug: an extra authproxy sidecar
In authenticated sessions, there is an `oauth2_proxy` sidecar in addition to a `authproxy` sidecar; anonymous sessions fail to create because there are two identical `authproxy` sidecards so the pod spec is invalid.

rokroskar:
fixes #954 

/deploy


aledegano (dismissed):

rokroskar:
> @rokroskar try to replace it with `ghcr.io/devcontainers-extra/features/poetry:2` The repository used here has been deprecated for a while iirc

That fixed it, thanks @leafty!

Issue #779, opened 2024-12-04
enhancement: propagate the annotations and labels from the CR to all child resources


olevski:
~This should be an option in the CRD. By default it should not propagate (so that we do not break the API since we currently do not propagate).~

~Actually it may be nicer to have 2 fields, one for annotations and one for labels.~

~When set all annotations and labels from the CR should be added to the children.~

EDIT:

~At the CRD we can add `Spec.Template.Metadata.Labels` and `Spec.Template.Metadata.Annotations`. The type for each can come from the k8s library.~

~In the rare case of collisions the original child annotation or label should prevail and not be replaced by the one from the CR. This is because the original child annotation or label can be used for example for auto scaling or similar things.~

~By default the template metadata and labels are empty and so no value is added to the children. Just like now.~

EDIT2:

Follow the same pattern as how a deployment or statefulset has a template field that is then used to add labels/annotations to the children (i.e. pods).

So we can add a `template` field to the AmaltheaSession spec and add the annotation and label field there. This is the same way as what Deployments and Statefulsets do. For AmaltheaSessions this should then propagate to the Statefulsets and to the pods.

```
spec:
  template:
    metadata:
      labels:
        labelName: someValue
     annotations:
       annotationName: someValue
```

olevski:
Also for the statefulset these should propagate all the way to the Pod.

leafty:
I suggest having allow lists of annotations/label domains, e.g. "allow renku.io/* annotations". This way there is much less chance of collision with k8s or other CRD annotations clashing.

leafty:
Case in point: things like `app.kubernetes.io/created-bypp.kubernetes.io/created-by` would most likely break controllers if propagated.

olevski:
I think I would rather actually just have annotations or labels that you have to specify that are just applied to all children. This way we dont have to do partial matches / searches across the keys as with something like `renku.io/*`. 

And it is also closer to how you can specify the pod annotations/labels from a deployment or statefulset.

leafty:
That works too

olevski:
For example a deployment has:

```
spec:
  template:
    metadata:
      labels:
        labelName: someValue
     annotations:
       annotationName: someValue
```

PR #964, opened 2025-07-03
feat: add children labels and annotations

olevski:
closes https://github.com/SwissDataScienceCenter/amalthea/issues/779

In the case of conflicts with selector labels or anything that is set on the children internally for amalthea  then the value set from amalthea internally will have precendece.

/deploy

### DIFF
--- a/internal/controller/children.go
+++ b/internal/controller/children.go
@@ -93,8 +93,8 @@ func (c ChildResource[T]) Reconcile(ctx context.Context, clnt client.Client, cr 
 				fallthrough
 			case amaltheadevv1alpha1.Always:
 				current.Spec = desired.Spec
-				current.ObjectMeta.Labels = desired.ObjectMeta.Labels
-				current.ObjectMeta.Annotations = desired.ObjectMeta.Annotations
+				current.Labels = desired.Labels
+				current.Annotations = desired.Annotations
 			default:
 				return fmt.Errorf("attempting to reconcile ingress with unknown stategy %s", strategy)
 			}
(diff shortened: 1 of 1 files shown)


api/v1alpha1/amaltheasession_children.go
```diff
@@ -641,3 +645,12 @@ func (as *AmaltheaSession) Secret() v1.Secret {
 	}
 	return secret
 }
+
+func (cr *AmaltheaSession) childLabels() map[string]string {
+	labels := map[string]string{}
+	maps.Copy(labels, cr.Spec.Template.Metadata.Labels)
+	// NOTE: stuff from selectorLabels will overwrite conflicts in labels (if there are any)
+	// This is the desired behaviour, we do not want to overwrite the selector labels.
+	maps.Copy(labels, selectorLabels(cr.Name))
```
sgaist, line 654: While that's what we want, should we have a warning printed so that admins can see something might be off if these labels are set ?

olevski: @sgaist I would rather address this in a followup PR. I am still unsure of where to put this warning. And I think I will add it to the status conditions as a dedicated condition.

api/v1alpha1/amaltheasession_types.go
```diff
@@ -123,6 +123,10 @@ type AmaltheaSessionSpec struct {
 	// +optional
 	// The name of the service account that should be used for the session Pod
 	ServiceAccountName string `json:"serviceAccountName,omitempty"`
+
+	// +optional
+	// Template for the fields that should be added to all children (and their children if applicable).
+	Template Template `json:"template,omitzero"`
```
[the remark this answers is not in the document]
olevski: I think there was some weird behaviour with omitempty. I am not sure.

api/v1alpha1/amaltheasession_types.go
```diff
@@ -551,6 +555,15 @@ type ReadinessProbe struct {
 	Type ReadinessProbeType `json:"type,omitempty"`
 }
 
+type Template struct {
+	Metadata TemplateMetadata `json:"metadata,omitzero"`
```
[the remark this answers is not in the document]
leafty: OK, I see, so we should have `omitzero` for structs 👍 .

leafty: I think I finally found what breaks with these changes:

```suggestion
	// +optional
	Metadata TemplateMetadata `json:"metadata,omitzero"`
```

Without this, the validator always expects to have a `metadata` field, which you can see fails the tests. If you add the optional, it should remove the `required` from the generated schema and that should make the CRD changes compatible with the existing custom resources.

leafty (commented):
Looks good to me, but I would merge only after we have a corresponding PR in `renku-data-services` which show that the template field works as intended.

PR #966, opened 2025-07-04
chore: fix upload of snyk image check reports

olevski:
For some reason snyk ignores the option that is supposed to make the file name be a specific value and not snyk.sarif.

I checked the docs and it seems that we are using the right flag that is supposed to give us a custom name for the sarif file. But the snyk action keeps producing only a `snyk.sarif` file.

### DIFF
--- a/.github/workflows/publish-chart.yaml
+++ b/.github/workflows/publish-chart.yaml
@@ -50,28 +50,27 @@ jobs:
         continue-on-error: true
         with:
           image: ${{ steps.publish-chart.outputs.amalthea-sessions-image}}
-          args: --file=./Dockerfile --severity-threshold=high --sarif-file-output=as.sarif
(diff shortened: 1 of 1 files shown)


olevski:
This fixes the publishing of the helm chart when we make a release.

olevski:
Here is the test git workflow failing and listing the files with the old code:
https://github.com/SwissDataScienceCenter/amalthea/actions/runs/16069374888/job/45350331743

running the docker image that does the scanning and checking its docs shows that the flag we are trying to use should work but I guess they have  a bug or some weird interaction with other flags.

```
--sarif-file-output=<OUTPUT_FILE_PATH>
    Save test output in SARIF format directly to the <OUTPUT_FILE_PATH> file, regardless of
    whether or not you use the --sarif option.
```

sambuc (approved):

PR #974, opened 2025-07-25
chore: update golang ci linter to 2.3.0

olevski:
Updates the linter verison.

Also addresses some new issues that arise. These new issues show up because the new version of the linter merged multiple separate linters into one larger linter. So now we are getting even more checks (that we did not have before).

None of them seem unreasonable and they were easy to fix.


leafty (approved):
:+1:

PR #977, opened 2025-08-06
chore: update deploy workflow for Azure

olevski:
/deploy


olevski:
The tests are failing for other reasons not because of this PR. I will submit a fix

PR #988, opened 2025-08-22
exp: use a dev name for sessions

leafty:
/deploy #notest renku=leafty/exp-dev-naming renku-data-services=leafty/exp-dev-naming


SalimKayal (approved):
LGTM

PR #998, opened 2025-09-04
chore: add chartpress-cleanup make target

leafty:


helm-chart/amalthea/values.yaml
```diff
@@ -88,7 +88,7 @@ image:
   repository: renku/amalthea
   pullPolicy: IfNotPresent
   # Overrides the image tag whose default is the chart appVersion.
-  tag: ""
+  tag: 0.1.0
```
olevski, line 91: I think this should either unset or latest and we should make the ci action deploy "latest" on merges to master. But having 0.1.0 as the default will result in a really old version. I suspect the last "default" is that old but I dont know that this improves anything. So I would just not touch it. Same applies to all the other tags in the values file.

Makefile
```diff
@@ -303,6 +304,10 @@ list-chartpress-images:
 	@rm chartpress.yaml
 	@mv chartpress.backup.yaml chartpress.yaml
 
+.PHONY: chartpress-cleanup
+chartpress-cleanup:
+	@chartpress --tag 0.1.0 --no-build
```
olevski, line 309: question: We can just run the chartpress command to reset. There is a specific flag for that. So why not do that? I.e. you can run `chartpress --reset`.

leafty: Done in f32f42bdafea58c794aa8eae23b758385cb58e1d.

PR #1016, opened 2025-09-30
fix: ensure cuda enabled images work on eiger with enroot

SalimKayal:


internal/remote/firecrest/session_script.sh
```diff
@@ -133,6 +133,12 @@ mkdir -p "${LOGS_DIR}"
 wstunnel=$(install_wstunnel)
 echo "wstunnel: ${wstunnel}"
 
+# Ensure NVIDIA_VISIBLE_DEVICES is set to void 
+# so that cuda enabled images work on eiger
+if !(nvidia-smi 2>&1 >/dev/null);then
```
sgaist, line 138: nitpick
```suggestion
if !(nvidia-smi 2>&1 >/dev/null); then
```

sgaist (approved):

PR #1019, opened 2025-09-30
build: use crdoc directly in actions

leafty:
`crdoc` is a go tool, no need to hide it inside some container image.

Using the `tool` directive allows us to directly manage `crdoc` and its version.

Tested by creating this PR -> https://github.com/SwissDataScienceCenter/amalthea/pull/1020


sgaist:
@leafty I think it was done like that because the original implementation of Amalthea was in Python and the tool was not as readily accessible as it is now that the code base has moved over to Go.

olevski (approved):

Issue #1014, opened 2025-09-29
idle sessions exceed the max idle time by many minutes before being removed


rokroskar:
Idle sessions do not get removed when reaching the idleness limit, but some time later (sometimes more than 10 minutes later).

olevski:
The current delay in amalthea 0.22.0 is 10 seconds and it is added as a delay to requeuing only when the session is fully running. In all other cases we dont put a delay.

What we can try is:
- EDIT: Look at the performance metrics - we started uploading some data to sentry and see if we can see what is causing large delays
- Requeue after  a long time (i.e. 30 mins) if the session is hibernated. If it is resumed an event will be queued immediately because the CR changed. We cannot set this to zero because we have a timer to delete the session if it is hibernated for too long.
- In all other keep the 10 sec delay. This is just to avoid the queue getting constantly hammered when there are lots of sessions.

PR #1032, opened 2025-10-22
fix: increase requeue for hibernated sessions

olevski:
/deploy

related to https://github.com/SwissDataScienceCenter/amalthea/issues/1014


bundle/manifests/amalthea.dev_amaltheasessions.yaml
```diff
@@ -5985,6 +5985,20 @@ spec:
                   - name
                   type: object
                 type: array
+              location:
```
olevski: The changes in this file are just leftovers from other PRs that are already merged. They CRDs in the helm chart already have this update. This is for deploying directly from manifests.

PR #1046, opened 2025-11-25
feat: Add `lastInteraction` property storing last seen user interaction

eikek:
For detecting when to hibernate a session, we use cpu threshold to detect an idle session. If a user explicitly proves to interact with the session, this timestamp is now stored as well. When it exists and it is after the `idleSince` time, it is used as the starting point for counting down the `MaxIdleTime`.

Removes the logic using last request time for detecting idleness.

Makes the previously hard-coded cpu threshold for idleness-checking configurable.

/deploy

### DIFF
--- a/internal/controller/children.go
+++ b/internal/controller/children.go
@@ -643,10 +643,7 @@ func (c ChildResourceUpdates) Status(
 		failingSince = metav1.Time{}
 	}
 
-	hibernationDate := metav1.Time{}
-	if pod != nil {
-		hibernationDate = calculateHibernationDate(log, pod.GetCreationTimestamp(), cr.Status, cr.Spec.Culling)
-	}
+	hibernationDate := calculateHibernationDate(log, cr.GetCreationTimestamp(), cr.Status, cr.Spec.Culling)
 
 	status := amaltheadevv1alpha1.AmaltheaSessionStatus{
 		Conditions:            Conditions(state, ctx, r, cr),


api/v1alpha1/amaltheasession_types.go
```diff
@@ -312,6 +312,16 @@ type Culling struct {
 	// Golang's time.ParseDuration is used to parse this, so values like 2h5min will work,
 	// valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".
 	MaxHibernatedDuration metav1.Duration `json:"maxHibernatedDuration,omitempty"`
+	// +optional
+	// +kubebuilder:validation:Format:=date-time
+	// +kubebuilder:validation:Type:=string
+	// A timestamp denoting the time when a user approved to have interacted with the session,
+	// preventing it from culling. This timestamp is used to restart MaxIdleDuration.
```
leafty: Let's rephrase this:


```suggestion
	// A timestamp denoting the time when a user has proven to have interacted with the session,
	// preventing it from culling.
	// This means that the following invariant is always true: `status.idleSince <= spec.culling.lastInteraction`.
```

internal/controller/cull.go
```diff
@@ -53,24 +55,63 @@ func updateHibernationState(ctx context.Context, r *AmaltheaSessionReconciler, a
 	return nil
 }
 
-func needsScaleDown(creationTimestamp metav1.Time, status amaltheadevv1alpha1.AmaltheaSessionStatus, culling amaltheadevv1alpha1.Culling) bool {
+func needsScaleDown(log logr.Logger, creationTimestamp metav1.Time, status amaltheadevv1alpha1.AmaltheaSessionStatus, culling amaltheadevv1alpha1.Culling) bool {
```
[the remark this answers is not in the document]
eikek: Yes, this was discussed a few times. In my opinion we should keep the meaning of `idle` as it is (like based only on measurements) and track the last user interaction separately. But we should take both things into account when deciding whether to auto shutdown a session. I agree that the current name `MaxIdleDuration` is not quite capturing the new decision process. We could perhaps use the proposed `willHibernateAtOrAfter` or `SessionTimeout` or something. So we can talk about the user visible action separate from how it is determined. OTOH I'm happy to do whatever you think is appropriate, just let me know 🙏🏼

internal/controller/cull.go
```diff
@@ -41,36 +44,170 @@ func updateHibernationState(ctx context.Context, r *AmaltheaSessionReconciler, a
...
+	} else {
+		logMsg += fmt.Sprint(" => ", result.Name, ":", result.Date)
+	}
+	log.Info("hibernationDate", "DecisionLog", strings.TrimSpace(logMsg))
+	return result.Date
+}
+
+func needsScaleDown(hibernationDate metav1.Time) bool {
+	now := time.Now()
+	return !hibernationDate.IsZero() && (hibernationDate.Time.Equal(now) || hibernationDate.Time.Before(now))
```
leafty: The case where equality would happen is vanishingly small, so we do not need to have it:

```suggestion
	return !hibernationDate.IsZero() && hibernationDate.Time.Before(now)
```

api/v1alpha1/amaltheasession_types.go
eikek: I updated the comment, hopefully the intention is clear.

internal/controller/children.go
```diff
@@ -643,6 +643,11 @@ func (c ChildResourceUpdates) Status(
 		failingSince = metav1.Time{}
 	}
 
+	hibernationDate := metav1.Time{}
+	if pod != nil {
+		hibernationDate = calculateHibernationDate(log, pod.GetCreationTimestamp(), cr.Status, cr.Spec.Culling)
+	}
```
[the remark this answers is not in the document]
eikek: Ok thanks, I just copied the previous code here. The `cr` is passed in as argument and has never been checked for being `nil` in this method. So I think I would then not do it somewhere in the middle of the method?

PR #1057, opened 2026-01-14
fix: Use HostToContainer in DC mounts

sambuc:
This is needed to support rebinding in case of the csi driver restarting.


leafty (changes requested):
You can use ptr.To() (see many uses in this code repository).

sambuc:
Not relevant anymore.

PR #1088, opened 2026-03-13
fix: handle default priority class and remote SA

sambuc:
Add flags to create:
 - a default resource quota priority class for renku user sessions
 - a service account with a role and role binding to allow a remote renku portal to manage Amalthea sessions.

### DIFF
--- a/helm-chart/amalthea-sessions/templates/priorityclass.yaml
+++ b/helm-chart/amalthea-sessions/templates/priorityclass.yaml
@@ -1,0 +1,12 @@
+{{- if .Values.deploy.priorityClass -}}
+---
+apiVersion: scheduling.k8s.io/v1
+description: Renku default resource quota priority class
+kind: PriorityClass
+metadata:
+  labels:
+    app: renku
+  name: renku-user-sessions-priority
+preemptionPolicy: Never
+value: 100
+{{- end }}
--- a/helm-chart/amalthea-sessions/templates/serviceaccount.yaml
+++ b/helm-chart/amalthea-sessions/templates/serviceaccount.yaml
@@ -1,3 +1,4 @@
+---
 apiVersion: v1
 kind: ServiceAccount
 metadata:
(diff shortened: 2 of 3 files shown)


helm-chart/amalthea-sessions/templates/priorityclass.yaml
```diff
@@ -0,0 +1,12 @@
+{{- if .Values.deploy.priorityClass -}}
+---
+apiVersion: scheduling.k8s.io/v1
+description: Renku default resource quota priority class
+kind: PriorityClass
+metadata:
+  labels:
+    app: renku
+  name: renku-user-sessions-priority
```
olevski, line 9: This is never referenced in the source code of amalthea. Why is it part of the helm chart? Amalthea does not make or manage priority classes.

helm-chart/amalthea-sessions/templates/serviceaccount.yaml
```diff
@@ -8,4 +9,108 @@ metadata:
...
-    {{- toYaml .Values.controllerManager.serviceAccount.annotations | nindent 4 }}
\ No newline at end of file
+    {{- toYaml .Values.controllerManager.serviceAccount.annotations | nindent 4 }}
+
+{{- if .Values.deploy.remoteServiceAccount }}
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: renku-remote-session-manager
```
olevski, line 19: This service account is not used anywhere in amalthea? Why have it here?

olevski: If this is the same "service account name" used in the cluster spec on the data service then that is used to assign a specific service account to the user session that is created. And in almost all cases the sessions do not need a dedicated service account. The only place we have needed this is on openshift where a dedicated service account is used to get a different security context constraint.

olevski (changes requested):
Hey Lionel we should discuss this. I am pretty sure we just don need any of the changes in here. In the worst case if we do need something we can document and ask admins to create them manually on the remote cluster outside of a helm chart.

aledegano:
> Hey Lionel we should discuss this. I am pretty sure we just don need any of the changes in here. In the worst case if we do need something we can document and ask admins to create them manually on the remote cluster outside of a helm chart.

Hey Tasko, I can give some context here: I have asked Lionel to add some of these manifests after I did the exercise of deploying on a Switch Cloud Kaas cluster (although the provider doesn't matter).
I was following the instructions that he [wrote ](https://github.com/SwissDataScienceCenter/renku/blob/feat/remote-clusters/docs/docs/20-admins/30-operation/50-remote.md) and remarking that there are a bit too many manual/out-of-band operations that should -if possible- be brought back into the chart.

Perhaps we could make everyone happy by having a toggle for this chart to select if it's deployed alongside Renku or as an agent/remote-operator?

sambuc:
I made the deployment of these two things based on conditions which are set to false by default.

The idea is that when an admin is OK with using the default value, they just set the conditions they want to true, or leave it as false, and create by hand the associated resources.

PR #1089, opened 2026-03-13
fix: allow adding parameters to OAuth2Proxy

sambuc:


helm-chart/amalthea-sessions/values.yaml
```diff
@@ -51,6 +51,11 @@ sidecars:
   image:
     repository: renku/sidecars
     tag: latest
+  oauth2Proxy:
+    # When deployed on an external cluster, it may be necessary to add
+    extraArgs:
+      []
+#      - "--cookie-samesite=none"
```
leafty, line 58: Comment: I'm not sure it's a good idea to turn off a web feature designed to prevent XSS.

sambuc:
not relevant anymore.

PR #1095, opened 2026-03-27
feat: Add non-interactive sessions

eikek:
A session can now be either interactive (the default) or non-interactive which is indicated by a new property in the CRD: `sessionType`. It is either `Interactive` or `NonInteractive`.

When a non-interactive session is requested, the session is not mapped to a stateful set, but to a job. Additionally, no `Service` and no `Ingress` is applied. The "hibernation" feature of a session is mapped to a suspended job.

The session spec should contain the two timeouts used for non-interactive sessions: `MaxAge` and `MaxHibernatedTime`. The former is pushed into the jobs `ActiveDeadlineSeconds` that defines the maximum runtime. If that limit is reached, the job is tried to terminate and the session goes into "Hibernated" state. From there the `MaxHibernatedTime` functions the same as for interactive sessions: once this is reached, the session and all related resources are removed. This is done by setting the `TTLSecondsAfterFinished` timeout on the Job spec. It is important, that the job lifetime (`ActiveDeadlineSeconds`) is not shorter than the amalthea session, otherwise the session will re-create the job when it is removed by kubernetes due to this timeout.

The overall amalthea status is derived from the associated Pod, which is assumed to be a single one. This is true for non-interactive sessions as well, as the config ensures only one pod running for the job. Then only the `GetPod` function is adopted to return the Pod of the Job in case of a non-interactive session.

/deploy


internal/controller/amaltheasession_controller.go
```diff
@@ -144,7 +144,14 @@ func (r *AmaltheaSessionReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
 	}
 
-	children, err := NewChildResources(amaltheasession, r.ClusterType)
+	var children ChildResources
+	switch amaltheasession.Spec.SessionType {
+	case amaltheadevv1alpha1.SessionTypeInteractive:
+		children, err = NewChildResources(amaltheasession, r.ClusterType)
+	case amaltheadevv1alpha1.SessionTypeNonInteractive:
+		children, err = NewJobChildResources(amaltheasession, r.ClusterType)
+	}
```
leafty: Comment: for readability, I would recommend having the following:

```go
children, err := NewChildResources(amaltheasession, r.ClusterType)

func NewChildResources(amaltheasession ...) {
  switch amaltheasession.Spec.SessionType {
	case amaltheadevv1alpha1.SessionTypeInteractive:
		return NewInteractiveChildResources(amaltheasession, r.ClusterType)
	case amaltheadevv1alpha1.SessionTypeNonInteractive:
		return NewNonInteractiveChildResources(amaltheasession, r.ClusterType)
	}
}
```

eikek: yes, makes sense, thanks!

api/v1alpha1/amaltheasession_children.go
```diff
@@ -95,45 +97,62 @@ func (cr *AmaltheaSession) SessionVolumes() ([]v1.Volume, []v1.VolumeMount) {
...
 	_, dsVols, dsVolMounts := cr.DataSources()
 	cloneInit := cr.cloneInit()
 	sessionVols, sessionMounts := cr.SessionVolumes()
+
+	auth, err := cr.auth()
+	if err != nil {
+		return nil, err
+	}
+
+	volumes := make([]v1.Volume, 0, len(sessionVols)+len(cloneInit.Volumes)+len(cr.Spec.ExtraVolumes)+len(dsVols)+len(auth.Volumes))
```
leafty: Note: I don't think it is particularly useful to "optimize" memory here.

api/v1alpha1/amaltheasession_children.go
```diff
@@ -95,45 +97,62 @@ func (cr *AmaltheaSession) SessionVolumes() ([]v1.Volume, []v1.VolumeMount) {
...
-	if cr.Spec.Hibernated {
-		replicas = 0
-	}
-
-	volumes := []v1.Volume{}
-	volumeMounts := []v1.VolumeMount{}
-	initContainers := []v1.Container{}
-	containers := []v1.Container{}
-
+func (cr *AmaltheaSession) Pod(cfg config.AmaltheaSessionConfiguration) (*v1.PodSpec, error) {
```
leafty, line 101: Why return a pointer to podspec? We can also pass by value.

api/v1alpha1/amaltheasession_children.go
```diff
@@ -158,6 +177,73 @@ func (cr *AmaltheaSession) StatefulSet(cfg config.AmaltheaSessionConfiguration)
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        cr.JobName(),
+			Namespace:   cr.Namespace,
+			Labels:      cr.childLabels(),
+			Annotations: cr.Spec.Template.Metadata.Annotations,
+		},
+		Spec: batchv1.JobSpec{
+			Parallelism:             &parallel,
+			Completions:             &parallel,
+			BackoffLimit:            &zero,
```
leafty: Easier to read if using `ptr.To` (https://pkg.go.dev/k8s.io/utils/ptr#To).

```suggestion
			Parallelism:             ptr.To(int32(1)),
			Completions:             ptr.To(int32(1)),
			BackoffLimit:            ptr.To(int32(0)),
```

api/v1alpha1/amaltheasession_children.go
```diff
@@ -395,21 +481,60 @@ func NewConditions() []AmaltheaSessionCondition {
...
+		selector := labels.Set{"job-name": cr.JobName()}.AsSelector()
+		podList := &v1.PodList{}
+		listOpts := &client.ListOptions{Namespace: cr.Namespace, LabelSelector: selector}
+		if err := clnt.List(ctx, podList, listOpts); err != nil {
+			return nil, err
+		}
+		if len(podList.Items) > 0 {
+			return &podList.Items[0], nil
+		}
+		return nil, nil
```
leafty: Should it be an error if the length is not exactly equal to 1?

config/samples/culling-configured.yaml
```diff
@@ -12,8 +12,8 @@ spec:
     readinessProbe:
       type: "none"
   culling:
-    maxAge: 10m
-    maxIdleDuration: 2m
+    maxAge: 5m
+    maxIdleDuration: 20s
```
leafty, line 16: ?

internal/controller/children.go
```diff
@@ -291,12 +294,127 @@ func (c ChildResource[T]) Reconcile(ctx context.Context, clnt client.Client, cr
...
+				if !(cr.Status.State == amaltheadevv1alpha1.Failed || cr.Status.State == amaltheadevv1alpha1.Hibernated) {
+					return nil
+				}
+				fallthrough
+			case amaltheadevv1alpha1.Always:
+				current.Spec.Template.Spec.Containers = desired.Spec.Template.Spec.Containers
+				current.Spec.Template.Spec.InitContainers = desired.Spec.Template.Spec.InitContainers
+				current.Spec.Template.Spec.Volumes = desired.Spec.Template.Spec.Volumes
+				// current.Spec.Selector = desired.Spec.Selector
+				// current.Spec.Template.Labels = desired.Spec.Template.Labels
```
leafty: ?

api/v1alpha1/amaltheasession_types.go
```diff
@@ -127,6 +151,20 @@ type AmaltheaSessionSpec struct {
 	// +optional
 	// Template for the fields that should be added to all children (and their children if applicable).
 	Template Template `json:"template,omitempty"`
+
+	// +optional
```
leafty: This should be non-optional and also immutable.


```suggestion
	// +kubebuilder:default:="Interactive"
	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="sesion type is immutable"
```

api/v1alpha1/amaltheasession_types.go
```diff
@@ -127,6 +151,20 @@ type AmaltheaSessionSpec struct {
...
+	SessionType SessionType `json:"sessionType,omitempty"`
+}
+
+// Return the sessionType or - if empty - the default SessionTypeInteractive.
+func (c *AmaltheaSessionSpec) GetSessionType() SessionType {
+	if c.SessionType == "" {
+		return SessionTypeInteractive
+	}
+	return c.SessionType
 }
```
leafty: Not necessary (see other comments, which force a default value).


```suggestion
```

api/v1alpha1/amaltheasession_types.go
```diff
@@ -26,6 +27,29 @@ import (
...
+}
+
+func (s *SessionType) UnmarshalJSON(data []byte) error {
+	var tmp string
+	if err := json.Unmarshal(data, &tmp); err != nil {
+		return err
+	}
+	*s = SessionType(tmp)
+	return nil
+}
```
leafty: This whole block should be moved below the definition of `AmaltheaSessionSpec`, most likely towards the end of the file.

This is to keep the root definitions at the top of the file.

api/v1alpha1/amaltheasession_types.go
```diff
@@ -26,6 +27,29 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// SessionType describes the type of session to be started. If none is specified, the
+// default SessionType will be "Interactive"
+// +enum
```
leafty:
```suggestion
// +kubebuilder:validation:Enum={Interactive,NonInteractive}
// SessionType describes the type of session to be started. If none is specified, the
// default SessionType will be "Interactive"
```

api/v1alpha1/amaltheasession_types.go
leafty: Not needed:


```suggestion
```

internal/controller/children.go
```diff
@@ -291,12 +294,127 @@ func (c ChildResource[T]) Reconcile(ctx context.Context, clnt client.Client, cr
...
+
+func NewChildResources(cr *amaltheadevv1alpha1.AmaltheaSession, cfg config.AmaltheaSessionConfiguration) (ChildResources, error) {
+	switch cr.Spec.GetSessionType() {
+	case amaltheadevv1alpha1.SessionTypeInteractive:
+		return NewInteractiveChildResources(cr, cfg)
+	case amaltheadevv1alpha1.SessionTypeNonInteractive:
+		return NewNonInteractiveChildResources(cr, cfg)
+	default:
+		return NewInteractiveChildResources(cr, cfg)
+	}
```
leafty:
```suggestion
	switch cr.Spec.GetSessionType() {
	case amaltheadevv1alpha1.SessionTypeNonInteractive:
		return NewNonInteractiveChildResources(cr, cfg)
	default:
		return NewInteractiveChildResources(cr, cfg)
	}
```

api/v1alpha1/amaltheasession_children.go
eikek: Yes, I agree. I got yelled at by some tool (from ci). 🤷🏼

config/samples/culling-configured.yaml
eikek: This was an attempt to make the tests successful. I noticed when looking in the commit history that I changed these values in the past. Back then, I didn't realize that this file is actually being used in a test. So these times matter for the tests - I think the name `samples/blabla.yaml` is not very telling. I thought it was just some example.

So here I changed it back to its original values. 😄

api/v1alpha1/amaltheasession_children.go
eikek: So i can return `nil` in case of an error. There is another function called inside that returns a pointer.

leafty: OK, though returning a zero-value in case of error is also how things are done with go.

api/v1alpha1/amaltheasession_children.go
eikek: Thanks!

config/samples/culling-configured.yaml
eikek: Sorry again. I thought these are new lines. But apparently I only changed the values back to the original ones.

api/v1alpha1/amaltheasession_children.go
eikek: Yes, I think this is not a correct state then. I can return an error in this case?

internal/controller/children.go
eikek: I got k8s errors when these were active, but I didn't understand the reason really. I need to rerun it and see again.

api/v1alpha1/amaltheasession_types.go
eikek: ok, thanks!

api/v1alpha1/amaltheasession_children.go
eikek: Ok, I can do it this way if you prefer? I don't have a strong opinion. I admit I'm still not friends with these zero values in go, they got me so many times now 😅 But `nil` is not better either…

api/v1alpha1/amaltheasession_types.go
eikek: Thanks!

api/v1alpha1/amaltheasession_types.go
eikek: Ok thank you! I can remember adding this after some problems, why is it not necessary?

internal/controller/children.go
eikek: 🤦🏼

internal/controller/children.go
eikek: yes, so these fields are immutable (from the error message). It works on other objects (like Service and StatefulSet). I removed these lines.

eikek:
Thank you @leafty for your sharp eyes. I addressed all comments except the two not marked as resolved, because I read it as non-blocking. But happy to change things, just let me know.

internal/controller/status.go
```diff
@@ -49,6 +49,31 @@ func podIsReady(pod *v1.Pod) bool {
 	return initCounts.Ok() && counts.Ok() && phaseOk
 }
 
+func podIsCompleted(pod *v1.Pod) bool {
+	if pod == nil || pod.GetDeletionTimestamp() != nil {
+		// A missing pod or a pod being deleted is not completed
+		return false
+	}
+
+	phaseOk := pod.Status.Phase == v1.PodSucceeded || pod.Status.Phase == v1.PodFailed
```
leafty: Why is phase ok true when the pod is failed?

internal/controller/status.go
```diff
@@ -49,6 +49,31 @@ func podIsReady(pod *v1.Pod) bool {
...
+	totalCnt := 0
+	completedCnt := 0
+	for _, container := range pod.Status.InitContainerStatuses {
```
leafty: Semantics: I don't think init containers should be counted here, only main containers should be counted so that we respect the k8s semantics of jobs and init containers with restart policy always.

internal/controller/children.go
```diff
@@ -366,23 +481,35 @@ func (c ChildResourceUpdates) IsRunning(pod *v1.Pod) bool {
...
-	case cr.Spec.Hibernated && c.StatefulSet.Manifest.Spec.Replicas != nil && *c.StatefulSet.Manifest.Spec.Replicas == 0:
+	case cr.Spec.Hibernated && c.StatefulSet.Manifest != nil && c.StatefulSet.Manifest.Spec.Replicas != nil && *c.StatefulSet.Manifest.Spec.Replicas == 0:
+		return amaltheadevv1alpha1.Hibernated, ""
+	case cr.Spec.Hibernated && c.Job.Manifest != nil && c.Job.Manifest.Spec.Suspend != nil && *c.Job.Manifest.Spec.Suspend:
 		return amaltheadevv1alpha1.Hibernated, ""
 	case msg != "":
 		return amaltheadevv1alpha1.Failed, msg
 	case c.IsRunning(pod):
 		return amaltheadevv1alpha1.Running, ""
+	case podIsCompleted(pod):
```
leafty, line 511: Why not use the job state here?

api/v1alpha1/amaltheasession_children.go
olevski: I think this is some linter getting updated and adding more checks with a new version. We can disable the check that triggered this error. I agree with Flora.

api/v1alpha1/amaltheasession_types.go
```diff
@@ -310,32 +317,37 @@ type Culling struct {
...
 	// Golang's time.ParseDuration is used to parse this, so values like 2h5min will work,
 	// valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".
 	MaxFailedDuration metav1.Duration `json:"maxFailedDuration,omitempty"`
 	// +kubebuilder:validation:Format:=duration
 	// How long can a session be in hibernated state before
-	// it gets completely deleted. A value of zero indicates that hibernated servers
+	// it gets completely deleted. A value of zero indicates that hibernated servers or jobs
 	// will not be automatically be deleted by Amalthea after a period of time.
 	// Golang's time.ParseDuration is used to parse this, so values like 2h5min will work,
 	// valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".
```
olevski, line 357:
```suggestion
	// +kubebuilder:validation:Format:=duration
	// How long can a session be in hibernated state before
	// it gets completely deleted. Or how long a Non-Interactive session 
        // will be kept around once it has completed.
        // A value of zero indicates that hibernated servers or jobs
	// will not be automatically be deleted by Amalthea after a period of time.
	// Golang's time.ParseDuration is used to parse this, so values like 2h5min will work,
	// valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".
```

api/v1alpha1/amaltheasession_children.go
```diff
@@ -95,45 +98,62 @@ func (cr *AmaltheaSession) SessionVolumes() ([]v1.Volume, []v1.VolumeMount) {
...
-
+func (cr *AmaltheaSession) Pod(cfg config.AmaltheaSessionConfiguration) (*v1.PodSpec, error) {
 	_, dsVols, dsVolMounts := cr.DataSources()
 	cloneInit := cr.cloneInit()
 	sessionVols, sessionMounts := cr.SessionVolumes()
+
+	auth, err := cr.auth()
+	if err != nil {
+		return nil, err
+	}
```
olevski: question: I dont know if this is specified elsewhere... but non-interactive sessions do not have ingresses and they do not expose ports. So there is no reason to add authentication to them. The authentication is just a proxy that checks credentials and forwards to the port or url the session exposes.

internal/controller/amaltheasession_controller.go
```diff
@@ -59,6 +59,7 @@ const secretCleanupFinalizerName = "amalthea.dev/secrets-finalizer"
 // +kubebuilder:rbac:groups=amalthea.dev,resources=amaltheasessions/finalizers,verbs=update
 // +kubebuilder:rbac:groups=core,resources=secrets,verbs=list;watch;delete;create;patch
 // +kubebuilder:rbac:groups=core,resources=pods,verbs=get;list;watch
+// +kubebuilder:rbac:groups=core,resources=jobs,verbs=get;list;create;watch;patch
```
olevski:
```suggestion
// +kubebuilder:rbac:groups=batch,resources=jobs,verbs=get;list;create;watch;patch
```

olevski: You can see an example of rbac for jobs controller here: https://book.kubebuilder.io/cronjob-tutorial/controller-implementation.html. It is in the 4th code text box.

internal/controller/children.go
eikek: I don't have the job at hand here. I thought to base it on the pod which is retrieved before either queried from the job name or looked up for a normal session. Do you suggest I should separate earlier, creating different `State` functions for normal sessions an non-interactive sessions?

internal/controller/status.go
eikek: It means "phase failed or succeeded are ok for meaning completed". I can rename it.

internal/controller/status.go
eikek: Ah yes, that makes sense. Thank you. But that would even apply to all pods, no?

api/v1alpha1/amaltheasession_children.go
olevski: We should also document in the crd spec that authentication is ignored for jobs.

olevski (changes requested):
Thanks Eike. It looks good. Just a few minor things.

internal/controller/status.go
leafty: But then we return status `Succeeded` when the job has failed. I don't know if this is going to be nice for users to see that their job is "Succeeded" and failed at the same time.

api/v1alpha1/amaltheasession_children.go
eikek: Ok, I then revert this optimization and try to mute the linter.

api/v1alpha1/amaltheasession_children.go
eikek: Ok, I then skip the auth container for jobs.

internal/controller/status.go
eikek: Ah this is not the case, this function is only meant to determine whether a pod is completed. The session is set to failed when an error message was found. This logic is untouched.

.golangci.yml
```diff
@@ -3,6 +3,8 @@ run:
   allow-parallel-runners: true
 linters:
   default: none
+  disable:
+    - prealloc
```
leafty: [non-blocking]
Was this failing or just printing a warning? We could use preallocation in some cases

api/v1alpha1/amaltheasession_children.go
```diff
@@ -95,45 +98,49 @@ func (cr *AmaltheaSession) SessionVolumes() ([]v1.Volume, []v1.VolumeMount) {
...
-	volumeMounts := []v1.VolumeMount{}
-	initContainers := []v1.Container{}
-	containers := []v1.Container{}
-
+func (cr *AmaltheaSession) Pod(cfg config.AmaltheaSessionConfiguration) (*v1.PodSpec, error) {
 	_, dsVols, dsVolMounts := cr.DataSources()
 	cloneInit := cr.cloneInit()
 	sessionVols, sessionMounts := cr.SessionVolumes()
+	isNonInteractive := cr.Spec.SessionType == SessionTypeNonInteractive
+	isInteractive := !isNonInteractive
```
leafty: [non-blocking] Maybe we could have a method on the `SessionType` type: `isInteractive := cr.Spec.SessionType.IsInteractive()`.

api/v1alpha1/amaltheasession_children.go
```diff
@@ -158,6 +165,70 @@ func (cr *AmaltheaSession) StatefulSet(cfg config.AmaltheaSessionConfiguration)
...
+		return batchv1.Job{}, err
+	}
+
+	pod.RestartPolicy = v1.RestartPolicyNever
+
+	// use MaxAge (amount of time until a session gets terminated) to set maximum job runtime
+	var activeTTL *int64
+	if cr.Spec.Culling.MaxAge.Seconds() > 0 {
+		vv := int64(cr.Spec.Culling.MaxAge.Seconds())
+		activeTTL = &vv
```
```suggestion
		activeTTL = ptr.To(int64(cr.Spec.Culling.MaxAge.Seconds()))
```

api/v1alpha1/amaltheasession_children.go
```diff
@@ -158,6 +165,70 @@ func (cr *AmaltheaSession) StatefulSet(cfg config.AmaltheaSessionConfiguration)
...
+	}
+
+	// use MaxHibernatedDuration (amount of time until a hibernated session is deleted) to set the time
+	// after which the job is removed after it has completed
+	var ttlFinish *int32
+	if cr.Spec.Culling.MaxHibernatedDuration.Seconds() > 0 {
+		vv := int32(cr.Spec.Culling.MaxHibernatedDuration.Seconds())
+		ttlFinish = &vv
```
```suggestion
		ttlFinish = ptr.To(int32(cr.Spec.Culling.MaxHibernatedDuration.Seconds()))
```

internal/controller/children.go
```diff
@@ -291,12 +294,123 @@ func (c ChildResource[T]) Reconcile(ctx context.Context, clnt client.Client, cr
...
+				fallthrough
+			case amaltheadevv1alpha1.Always:
+				current.Spec.Template.Spec.Containers = desired.Spec.Template.Spec.Containers
+				current.Spec.Template.Spec.InitContainers = desired.Spec.Template.Spec.InitContainers
+				current.Spec.Template.Spec.Volumes = desired.Spec.Template.Spec.Volumes
+				current.Labels = desired.Labels
+				current.Annotations = desired.Annotations
+				current.Spec.Template.Annotations = desired.Spec.Template.Annotations
+			default:
+				return fmt.Errorf("attempting to reconcile ingress with unknown stategy %s", strategy)
```
leafty: ingress?

leafty (commented):
Minor things:

api/v1alpha1/amaltheasession_types.go
eikek: I copied over this comment, only removeing the duplicate word `be` in _"will not **be** automatically **be** deleted"_

.golangci.yml
eikek: It was failing. But I didn't spent a lot of time diagnosing this. I will look that I can disable it for selected lines.

eikek: It was also reporting on code I didn't touch in this PR. Probably some linter update that added some new features. That's why I then thought to disable it for the entire repo. Now I added to all lines `//nolint:prealloc` to disable the error.

api/v1alpha1/amaltheasession_children.go
eikek: Ah thanks!

olevski (approved):

PR #1100, opened 2026-04-01
feat: use self-minted tokens for sessions

leafty:
Implement using internal authentication tokens from renku-data-services in `git-proxy` and in `remote-session-controller`.

See: https://github.com/SwissDataScienceCenter/renku-data-services/pull/1249

Note: code for handling token authentication in sidecars can be consolidated, will try to make it nicer in the future.


olevski (approved):
Thanks Flora. Looks good.

PR #1106, opened 2026-04-13
fix: allow overriding https for session ingress

olevski:
Allows enabling https for remote clusters.

### DIFF
--- a/api/v1alpha1/amaltheasession_children.go
+++ b/api/v1alpha1/amaltheasession_children.go
@@ -611,6 +611,10 @@ func getUseNoneSameSiteSessionCookie() bool {
 	return strings.ToLower(os.Getenv("USE_NONE_SAME_SITE_SESSION_COOKIE")) == "true"
 }
 
+func getHttpsSessionIngress() bool {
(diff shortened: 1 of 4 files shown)


leafty (changes requested):
Why is this needed when we have https://github.com/SwissDataScienceCenter/amalthea/pull/1104?

olevski:
> Why is this needed when we have https://github.com/SwissDataScienceCenter/amalthea/pull/1104?

Because it is easier to rollout or apply this to remote clusters. We may remove this option from the helm chart when we add proper support for this in the data service in the configuration for remote clusters. But for now I really need to get the meteo swiss remote cluster unstuck and this enables that.

leafty:
> > Why is this needed when we have #1104?
> 
> Because it is easier to rollout or apply this to remote clusters. We may remove this option from the helm chart when we add proper support for this in the data service in the configuration for remote clusters. But for now I really need to get the meteo swiss remote cluster unstuck and this enables that.

Can we develop outside of `main` then? I don't think you need to release components to do remote cluster development. This looks unprofessional.

leafty (approved):

PR #1109, opened 2026-04-17
build: upgrade go to v1.26.2

leafty:
Note: some dependency updates require newer versions of `go`.

/deploy #notest


sgaist:
The move to 1.26 has been done in #1171
Closing

PR #1112, opened 2026-04-20
feat: rewrite images

leafty:
Add a new configuration value `imageRewriteRules` in `helm-chart/amalthea-sessions/values.yaml` which allows session images to be rewritten.

This feature allows amalthea admins to configure rewrite rules in order to use a container image pull-through cache, for example with harbor: https://goharbor.io/docs/2.14.0/administration/configure-proxy-cache/.

/deploy extra-values=amalthea-sessions.imageRewriteRules[0].source_prefix=docker.io,amalthea-sessions.imageRewriteRules[0].target_prefix=harbor.renkulab.io/dockerhub_cache


internal/controller/config/image_rewriter.go
```diff
@@ -0,0 +1,90 @@
...
+	"encoding/json"
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/distribution/reference"
+)
+
+// ImageRewriter rewrites container image references
+type ImageRewriter interface {
```
sgaist: Should we rather use ImagePrefixRewriter ?

ImageRewriter sounds like we are going to do something about the image itself rather than the field content.

internal/controller/config/image_rewriter.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+// rewrite the domain and path part of a container image reference according to the rule
+func (rule *imageRewriteRule) rewrite(domainAndPath string) (match bool, result string) {
+	after, found := strings.CutPrefix(domainAndPath, rule.SourcePrefix)
+	if !found || !strings.HasPrefix(after, "/") {
+		return false, domainAndPath
+	}
+	return true, rule.TargetPrefix + after
+}
```
sgaist: I think go is usually written the other way around, having the boolean indicating a form of success as the second variable.
```suggestion
func (rule *imageRewriteRule) rewrite(domainAndPath string) (result string, match bool) {
	after, found := strings.CutPrefix(domainAndPath, rule.SourcePrefix)
	if !found || !strings.HasPrefix(after, "/") {
		return domainAndPath, false
	}
	return rule.TargetPrefix + after, true
}
```

internal/controller/config/image_rewriter.go
```diff
@@ -0,0 +1,90 @@
...
+	path := reference.Path(named)
+	domainAndPath := fmt.Sprintf("%s/%s", domain, path)
+	tagged, isTagged := named.(reference.Tagged)
+	digested, isDigested := named.(reference.Digested)
+	for _, rule := range r.rules {
+		// Skip malformed rules
+		if rule.SourcePrefix == "" || rule.TargetPrefix == "" {
+			continue
+		}
+		match, result := rule.rewrite(domainAndPath)
```
sgaist: Based on my comment above:
```suggestion
		result, match := rule.rewrite(domainAndPath)
```

internal/controller/config/image_rewriter_test.go
```diff
@@ -0,0 +1,117 @@
...
+	rule := imageRewriteRule{
+		SourcePrefix: "example.org",
+		TargetPrefix: "my-cache.com/prefix",
+	}
+
+	for _, testCase := range cases {
+		name := strings.ReplaceAll(testCase.input, "/", "_")
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+			match, result := rule.rewrite(testCase.input)
```
sgaist: Based on my comment above
```suggestion
			result, match := rule.rewrite(testCase.input)
```

sgaist (changes requested):
Implementation looks good

Just some question and suggestions but nothing I would deem blocking.

internal/controller/config/image_rewriter.go
leafty: The interface name and method are intentionally generic: to the consumer, it is just asking to rewrite the image. Nothing says that the rewrite will be about prefixes. Someone could add other rewriters later, e.g. "pyhton" -> "python" because a malicious image is parked there.

sgaist: Understood

Then what about `ImageFieldRewriter` for the interface ?
I feel there's a need to be a bit more precise while staying generic.

Note that if malicious images are a specific concern, I think tools like [Kyverno](https://kyverno.io) might be a better choice that doing it at the Amalthea level.

internal/controller/config/configuration.go
```diff
@@ -0,0 +1,23 @@
+package config
+
+import (
+	"k8s.io/client-go/rest"
+)
+
+type AmaltheaSessionConfiguration struct {
+	ClusterType   ClusterType
+	ImageRewriter ImageFieldRewriter
```
sgaist:
```suggestion
	ImageFieldRewriter ImageFieldRewriter
```

internal/controller/config/configuration.go
```diff
@@ -0,0 +1,23 @@
...
+func GetAmaltheaSessionConfiguration(config *rest.Config) (cfg AmaltheaSessionConfiguration, err error) {
+	clusterType, err := DetectClusterType(config)
+	if err != nil {
+		return AmaltheaSessionConfiguration{}, err
+	}
+	imageRewriter, err := GetImageFieldRewriter()
+	if err != nil {
+		return AmaltheaSessionConfiguration{}, err
+	}
+	return AmaltheaSessionConfiguration{ClusterType: clusterType, ImageRewriter: imageRewriter}, nil
```
sgaist:
```suggestion
	imageFieldRewriter, err := GetImageFieldRewriter()
	if err != nil {
		return AmaltheaSessionConfiguration{}, err
	}
	return AmaltheaSessionConfiguration{ClusterType: clusterType, ImageFieldRewriter: imageFieldRewriter}, nil
```

api/v1alpha1/amaltheasession_children.go
```diff
@@ -741,21 +742,36 @@ func makeTunnelSecret(length int) (string, error) {
...
 }
 
 // sessionContainer returns the main session container
-func (cr *AmaltheaSession) sessionContainerLocal(volumeMounts []v1.VolumeMount) v1.Container {
+func (cr *AmaltheaSession) sessionContainerLocal(volumeMounts []v1.VolumeMount, config config.AmaltheaSessionConfiguration) v1.Container {
 	session := cr.Spec.Session
+	// Rewrite image if no pull secrets are configured
+	image := session.Image
+	if len(cr.Spec.ImagePullSecrets) == 0 && config.ImageRewriter != nil {
+		newImage, err := config.ImageRewriter.Rewrite(image)
```
sgaist:
```suggestion
		newImage, err := config.ImageFieldRewriter.Rewrite(image)
```

helm-chart/amalthea-sessions/templates/deployment.yaml
```diff
@@ -47,6 +47,10 @@ spec:
           value: {{ .Values.useNoneSameSiteSessionCookie | quote }}
         - name: HTTPS_SESSION_INGRESS
           value: {{ .Values.httpsSessionIngress | quote }}
+        {{- with .Values.imageRewriteRules }}
+        - name: AMALTHEA_IMAGE_REWRITE_RULES
```
sgaist, line 51: Some food for thoughts: what about using a file for the rules ?
An environment variable has a size limit. While it's not exactly small, I think it's worth considering long term.

sgaist (changes requested):
Just some naming consistency suggestions otherwise things look good !

aledegano:
I just want to point out that once we deploy this our Harbor becomes a critical POF.
I think that's okay and we should be able to reliably keep it alive (it's already deployed in HA mode), but it's worth remembering.
And yes, we can switch off the feature in case Harbor does not cooperate...

leafty:
> I just want to point out that once we deploy this our Harbor becomes a critical POF. I think that's okay and we should be able to reliably keep it alive (it's already deployed in HA mode), but it's worth remembering. And yes, we can switch off the feature in case Harbor does not cooperate...

1. This feature should also work with other pull-through caches (e.g. if we use the Azure Container Registry).
2. Yes, if there is an issue with harbor, simply upgrade the helm release with the rewrite rules removed.

helm-chart/amalthea-sessions/templates/deployment.yaml
leafty: We do not use a configmap at the moment, so we should do that later. It would be nice to refactor the reconciler so that we use the best practices from `cobra` and `viper` instead of having a mess of command line arguments and env vars.

api/v1alpha1/amaltheasession_children.go
```diff
@@ -741,21 +742,36 @@ func makeTunnelSecret(length int) (string, error) {
...
 	return cr.sessionContainerRemote(volumeMounts)
 }
 
 // sessionContainer returns the main session container
-func (cr *AmaltheaSession) sessionContainerLocal(volumeMounts []v1.VolumeMount) v1.Container {
+func (cr *AmaltheaSession) sessionContainerLocal(volumeMounts []v1.VolumeMount, config config.AmaltheaSessionConfiguration) v1.Container {
 	session := cr.Spec.Session
+	// Rewrite image if no pull secrets are configured
+	image := session.Image
+	if len(cr.Spec.ImagePullSecrets) == 0 && config.ImageRewriter != nil {
```
sgaist: Sorry, I missed that one
```suggestion
	if len(cr.Spec.ImagePullSecrets) == 0 && config.ImageFieldRewriter != nil {
```

sgaist (changes requested):

sgaist (approved):

PR #1118, opened 2026-04-30
fix: remove debugging printing in git helper

### LINKED ISSUE #1117: Failing git commands leak env to stderr
If a git command fails, the full contents of env vars may be printed to `git-clone` logs, including secrets (access tokens, etc.). This is not desirable.

This seems to be a direct effect of these lines in `env-pass.sh`.

https://github.com/SwissDataScienceCenter/amalthea/blob/aa94b0b2839ab3f5d9494fb9b208eff20ff5d1a5/env-pass.sh#L6-L7

@sgaist Do we print the environment just for debugging?

sgaist:
## Describe your changes

Removed environment printing from git helper script.

It was used to help debug the functionnality.

## Tickets and links

Fixes #1117

### DIFF
--- a/env-pass.sh
+++ b/env-pass.sh
@@ -3,9 +3,6 @@
 set -eu
 set -o pipefail
 
-# Redirect to stderr so as to not interfere with the stdout needed for username/password prompts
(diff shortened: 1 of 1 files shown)


leafty (approved):
LGTM

PR #1120, opened 2026-05-08
fix: safe default for SessionType if the CRD is not up-to-date

leafty:
This change makes amalthea work when the installed CRD is pre-job (no `SessionType` field).

/deploy


eikek (approved):
Thank you

PR #1123, opened 2026-05-19
fix: session_script.sh: magic line has to be first

sambuc:
The magic line/interpreter line has to be the first of the file, otherwise it is simply ignored.

### DIFF
--- a/internal/remote/firecrest/session_script.sh
+++ b/internal/remote/firecrest/session_script.sh
@@ -1,10 +1,11 @@
+#!/bin/bash
+
 # NOTE FOR AMALTHEA MAINTAINERS:
 #   This script contains template strings in the following form:
 #     `#{{NAME}}`
 #   These strings should be added or removed according to code changes
 #   in the remote session controller.
 # END NOTE
-#!/bin/bash
 #{{SBATCH_DIRECTIVES_PLACEHOLDER}}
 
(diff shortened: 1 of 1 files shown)


leafty (changes requested):
???
The comments are removed before submission. This script is not meant to be executed "AS IS", which is the point of the NOTE FOR AMALTHEA MAINTAINERS at the top.

See templating code: https://github.com/SwissDataScienceCenter/amalthea/blob/b9823e1b9068d16cc21687464528bee717e8a230/internal/remote/firecrest/controller.go#L602-L674

internal/remote/firecrest/session_script.sh
```diff
@@ -1,10 +1,11 @@
+#!/bin/bash
+
 # NOTE FOR AMALTHEA MAINTAINERS:
```
leafty, line 2: I don't want empty lines before the SBATCH directives.

```suggestion
#!/bin/bash
# NOTE FOR AMALTHEA MAINTAINERS:
```

leafty (commented):
This should be tested on `staging`.

sambuc:
This is currently deployed on staging via https://github.com/SwissDataScienceCenter/renku/pull/4261

leafty (approved):
LGTM, checked rendered job script.

PR #1124, opened 2026-05-20
fix: Handle all firecrest FS types

sambuc:
* Simplify the logic, as we only want to handle specifically the "Users" fs type for now.
* Check secretPath and add it only if the string is not empty.

### DIFF
--- a/internal/remote/firecrest/controller.go
+++ b/internal/remote/firecrest/controller.go
@@ -635,40 +635,30 @@ func addSessionMountsToScript(sessionScript string, fileSystems *[]FileSystem, s
 	if fileSystems == nil {
 		return strings.Replace(sessionScript, "#{{SESSION_MOUNTS_PLACEHOLDER}}", "", 1)
 	}
-	// Collect file systems we want to mount
-	var home *FileSystem
-	scratches, stores := []*FileSystem{}, []*FileSystem{}
+
+	// Collect file systems we want to mount as "SRC:DST[:FLAG]" strings
+	var mounts []string
 	for _, fs := range *fileSystems {
 		switch fs.DataType {
-		case Scratch:
-			scratches = append(scratches, &fs)
-		case Store:
-			stores = append(stores, &fs)
 		case Users:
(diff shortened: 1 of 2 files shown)


internal/remote/firecrest/controller.go
```diff
@@ -635,40 +635,30 @@ func addSessionMountsToScript(sessionScript string, fileSystems *[]FileSystem, s
...
-	}
-	for _, store := range stores {
-		mounts = append(mounts, store.Path)
-	}
-	// TODO: Try to mount home at its location (need to handle ~/.bashrc)
-	// TODO: Alternatively, copy the contents in the container
-	if home != nil {
-		mounts = append(mounts, fmt.Sprintf("%s:/home%s:ro", home.Path, home.Path))
+	// Add the secrets mount
+	if len(secretsPath) > 0 {
```
leafty: Effective go:


```suggestion
	if secretsPath != "" {
```

internal/remote/firecrest/controller.go
```diff
@@ -635,40 +635,30 @@ func addSessionMountsToScript(sessionScript string, fileSystems *[]FileSystem, s
...
 		switch fs.DataType {
-		case Scratch:
-			scratches = append(scratches, &fs)
-		case Store:
-			stores = append(stores, &fs)
 		case Users:
-			home = &fs
+			// TODO: Try to mount home at its location (need to handle ~/.bashrc)
+			// TODO: Alternatively, copy the contents in the container
+			mounts = append(mounts, fmt.Sprintf("%s/${USER}:/home%s/${USER}:ro", fs.Path, fs.Path))
```
leafty: Please don't interpolate `${USER}`. This is potentially breaking.


```suggestion
			mounts = append(mounts, fmt.Sprintf("%s:/home%s:ro", fs.Path, fs.Path))
```

sambuc: I made a new commit to also revert the unit test.

leafty (approved):

PR #1128, opened 2026-06-03
fix: properly clean up finished jobs and return failure when max-age exceeded

eikek:
- Since finished jobs are actively deleted by amalthea, there is no need to specify `ttlSecondsAfterFinished` for a job. It will only introduce a race condition
- Once a job is done (success or failure), amalthea will remove it after `maxHibernatedDuration` is passed
- A finished job must not be touched in the reconcile loop, otherwise it will be restarted
- If a job exceeds its `MaxAge` time, k8s will terminate the pod and the job goes into failed state. This state is now properly propagated to the amalthea session status
- in `cull.go` the jobs are removed from being "hibernated", as this only applies to interactive sessions. Jobs are being removed by k8s automatically due to its `activeDeadlineSeconds` setting

/deploy

### DIFF
--- a/internal/controller/cull.go
+++ b/internal/controller/cull.go
@@ -30,10 +30,15 @@ import (
 )
 
 func updateHibernationState(ctx context.Context, r *AmaltheaSessionReconciler, amaltheasession *amaltheadevv1alpha1.AmaltheaSession) error {
+	if amaltheasession.Spec.SessionType == amaltheadevv1alpha1.SessionTypeNonInteractive {
+		// non-interactive sessions (jobs) never get auto-scaled down by amalthea;
+		// this is managed by k8s via the activeDeadlineSeconds setting for jobs
+		return nil
+	}
+
 	status := amaltheasession.Status
 	log := log.FromContext(ctx)
-	interactive := amaltheasession.Spec.SessionType == amaltheadevv1alpha1.SessionTypeInteractive
-	if !amaltheasession.Spec.Hibernated && interactive { // hibernating only applies to interactive sessions, jobs are managed for that by k8s
+	if !amaltheasession.Spec.Hibernated {
 		hibernationDate := status.WillHibernateAt
 		if status.State != amaltheadevv1alpha1.Hibernated && needsScaleDown(hibernationDate) {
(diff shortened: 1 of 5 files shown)


internal/controller/cull.go
```diff
@@ -32,15 +32,16 @@ import (
 func updateHibernationState(ctx context.Context, r *AmaltheaSessionReconciler, amaltheasession *amaltheadevv1alpha1.AmaltheaSession) error {
```
leafty, line 32: Suggestion: short-circuit at the top of the method

```go
if amaltheasession.Spec.SessionType == amaltheadevv1alpha1.SessionTypeNonInteractive {
  // non-interactive sessions (jobs) never get auto-scaled down by amalthea
  return nil;
}
```

internal/controller/cull.go
```diff
@@ -32,15 +32,16 @@ import (
...
 	status := amaltheasession.Status
 	log := log.FromContext(ctx)
-	if !amaltheasession.Spec.Hibernated {
+	interactive := amaltheasession.Spec.SessionType == amaltheadevv1alpha1.SessionTypeInteractive
```
leafty: I think this is unsafe (if an old custom resource is still present).
You can do `== amaltheadevv1alpha1.SessionTypeNonInteractive` or `!= amaltheadevv1alpha1.SessionTypeNonInteractive`.

internal/controller/cull.go
```diff
@@ -32,15 +32,16 @@ import (
...
+	if !amaltheasession.Spec.Hibernated && interactive { // hibernating only applies to interactive sessions, jobs are managed for that by k8s
```
leafty: Note: the comment may be misleading: amalthea does not auto-scale down. `Hibernating: true` is still accepted from the spec.

leafty (commented):
Looks OK, some small comments below.

internal/controller/cull.go
eikek: Ah right, thanks!!

leafty (approved):

olevski (approved):

PR #1130, opened 2026-06-10
fix: Allow setting git proxy variables

sambuc:
### PR Stack
* 🐸 #1130
* #1131
* #1132
* #1133
* #1134


internal/remote/firecrest/session_script.sh
```diff
@@ -9,8 +9,8 @@
 
 set -e -o pipefail
 
-GIT_PROXY_WAIT_SLEEP_SECONDS=10
-GIT_PROXY_WAIT_RETRIES=10
+: ${GIT_PROXY_WAIT_SLEEP_SECONDS:=10}
+: ${GIT_PROXY_WAIT_RETRIES:=10}
```
leafty, line 13: Can we use the `VAR=${VAR:=default}` syntax? I am not sure that most people in `@renku-devs` understand this one.

sambuc: I find the argument that we should not use a language construct / keyword because someone might not know it to be is extremely weak.

Now the colon (`:`) built-in operator is just that, an operator that can be used in different use cases, including this one.

I **strongly** insist to use what is in the patch, because:
 - this is less error prone (2x less chances of making a typo/copy-paste error)
 - This streamlines environment variable handling in a simple and robust construct, easy to understand where you can focus on what `VAR_NAME`, how (`:=`, `=`, `:?` , etc…) and the value/error message

leafty (approved):

PR #1132, opened 2026-06-10
chore: Allow setting the wstunnel version

sambuc:
### PR Stack
* #1130
* #1131
* 🐸 #1132
* #1133
* #1134

### DIFF
--- a/internal/remote/firecrest/session_script.sh
+++ b/internal/remote/firecrest/session_script.sh
@@ -16,9 +16,9 @@ set -e -o pipefail
 
 : ${WSTUNNEL_VERSION_x86_64:="10.4.4"}
 : ${WSTUNNEL_VERSION_aarch64:="10.1.10"}
-: ${d:="WSTUNNEL_VERSION_${ARCH}"}
-: ${WSTUNNEL_VERSION:="${!d}"}
-unset d
+: ${varname:="WSTUNNEL_VERSION_${ARCH}"}
+: ${WSTUNNEL_VERSION:="${!varname}"}
+unset varname
 
 # Installs rclone
 #
--- a/internal/remote/firecrest/session_script.sh
+++ b/internal/remote/firecrest/session_script.sh
@@ -9,9 +9,16 @@
 
 set -e -o pipefail
 
+: ${ARCH:=$(uname -m)}
 : ${GIT_PROXY_WAIT_SLEEP_SECONDS:=10}
 : ${GIT_PROXY_WAIT_RETRIES:=10}
 : ${RCLONE_VERSION:="1.70.2"}
+
+: ${WSTUNNEL_VERSION_x86_64:="10.4.4"}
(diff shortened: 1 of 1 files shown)


internal/remote/firecrest/session_script.sh
```diff
@@ -9,10 +9,17 @@
...
 set -e -o pipefail
 
+: ${ARCH:=$(uname -m)}
 : ${GIT_PROXY_WAIT_SLEEP_SECONDS:=10}
 : ${GIT_PROXY_WAIT_RETRIES:=10}
 : ${RCLONE_VERSION:="1.70.2"}
 
+: ${WSTUNNEL_VERSION_x86_64:="10.4.4"}
+: ${WSTUNNEL_VERSION_aarch64:="10.1.10"}
+: ${d:="WSTUNNEL_VERSION_${ARCH}"}
```
leafty: uh can we avoid creating bash variable with just one character?
We can have a small bash block with `if ... else` if needed to define the wstunnel version.

sambuc: Sorry about that, I have fixed it by using a proper variable name.

internal/remote/firecrest/session_script.sh
```diff
@@ -9,10 +9,17 @@
...
+: ${ARCH:=$(uname -m)}
 : ${GIT_PROXY_WAIT_SLEEP_SECONDS:=10}
 : ${GIT_PROXY_WAIT_RETRIES:=10}
 : ${RCLONE_VERSION:="1.70.2"}
 
+: ${WSTUNNEL_VERSION_x86_64:="10.4.4"}
+: ${WSTUNNEL_VERSION_aarch64:="10.1.10"}
+: ${varname:="WSTUNNEL_VERSION_${ARCH}"}
+: ${WSTUNNEL_VERSION:="${!varname}"}
+unset varname
```
leafty, line 21: This is not a better version than the existing one: we do not print a warning anymore.

leafty (commented):
As a general comment: maybe you can test if the latest version ([`v10.5.5`](https://github.com/erebe/wstunnel/releases/tag/v10.5.5)) is not segfaulting on grace hopper nodes. `wstunnel` may now be compiled with support for larger memory pages (https://github.com/erebe/wstunnel/blob/main/Cross.toml).

sambuc:
I made a mistake and pushed this on the rclone branch.

I'll fix the other branch as well.

sambuc:
> As a general comment: maybe you can test if the latest version ([`v10.5.5`](https://github.com/erebe/wstunnel/releases/tag/v10.5.5)) is not segfaulting on grace hopper nodes. `wstunnel` may now be compiled with support for larger memory pages (https://github.com/erebe/wstunnel/blob/main/Cross.toml).

As discussed, I have cloned the original repository and made a manual release of wstunnel with the build options we need.

I then added a fix to point to that release in the script and simplify things to only use a single version variable, like rclone

PR #1133, opened 2026-06-10
chore: Refactor URL generation

sambuc:
### PR Stack
* #1130
* #1131
* #1132
* 🐸 #1133
* #1134


internal/remote/firecrest/session_script.sh
```diff
@@ -20,13 +20,27 @@ set -e -o pipefail
 : ${WSTUNNEL_VERSION:="${!d}"}
 unset d
 
+case ${ARCH} in
+    "x86_64")
+        gh_arch=amd64
```
leafty, line 20: Why `gh_arch` and not `GH_ARCH`?

sambuc: Usually (exported) environment variables are in all capitals, while local (to the script/function) variables are in lowercase.

leafty: All other variables are in uppercase. This variable is not different from `WSTUNNEL_VERSION`.

sambuc: `gh_arch` is a local variable, I introduced it only to avoid having two `case … esac` blocks which do the same thing.

Like many other variables in this script I have set it in lower case.

leafty (approved):

leafty (approved):

PR #1135, opened 2026-06-10
feat: wire manifest readinessProbe to remote session controller

SalimKayal:
## Summary

The AmaltheaSession CRD already exposes spec.session.readinessProbe, spec.session.port, and spec.session.urlPath. These fields were previously only honored for local sessions (through Kubernetes container probes). This PR forwards the same values into the remote session controller sidecar so its /ready HTTP endpoint reports readiness based on whether the actual remote session is reachable.

## Motivation and Context

For remote sessions managed via FirecREST / RunAI / Slurm, the controller pod was always reporting ready immediately, even if the IDE on the remote compute node had not started yet. We need the Kubernetes readiness state to reflect real user-facing session availability so that ingress and service routing only start once the session is actually usable.

## Changes

*   Injects RSC_SESSION_PORT, RSC_SESSION_URL_PATH, and RSC_READINESS_PROBE_TYPE as environment variables into the remote session controller container (sessionContainerRemote).
*   The sidecar's /ready handler now interprets these values:
    *   none — immediately returns 200 OK.
    *   tcp — attempts a TCP dial to 127.0.0.1:<RSC_SESSION_PORT> (the local wstunnel endpoint); returns 200 on success, 503 otherwise.
    *   http — performs an HTTP GET to 127.0.0.1:<RSC_SESSION_PORT><RSC_SESSION_URL_PATH>; returns 200 on 2xx/3xx responses, 503 otherwise.
*   Fully backward compatible: an empty or unknown probe type, as well as port 0, fall back to the previous behavior of returning 200 OK.

**Tests**

*   Added TestConfigNewFlags covering CLI flags, environment variables, and validation of the readiness probe type.
*   Added TestReadyEndpoint covering all probe modes and edge cases: open/closed TCP ports, HTTP 200/302/500, connection refused, and port-0 fallback.

**Compatibility**

*   No CRD changes required — the fields already exist in amaltheasession_types.go.
*   Existing remote sessions without these values continue to behave exactly as before.

## Notes

- Companion PR: https://github.com/SwissDataScienceCenter/renku-data-services/pull/1350 selects the probe type per session; this PR acts on it in the sidecar. Merge/release together.

### DIFF
--- a/internal/remote/config/config.go
+++ b/internal/remote/config/config.go
@@ -143,6 +143,13 @@ func GetConfig() (cfg RemoteSessionControllerConfig, err error) {
 }
 
 func (cfg *RemoteSessionControllerConfig) Validate() error {
+	if cfg.ReadinessProbeType != "" &&
+		cfg.ReadinessProbeType != string(amaltheadevv1alpha1.None) &&
+		cfg.ReadinessProbeType != string(amaltheadevv1alpha1.TCP) &&
+		cfg.ReadinessProbeType != string(amaltheadevv1alpha1.HTTP) {
+		return fmt.Errorf("invalid readiness probe type: %s", cfg.ReadinessProbeType)
+	}
+
 	// FireCREST has priority over Runai
 	cfg.RemoteKind = RemoteKindFirecrest
 	firecrestConfigErr := cfg.Firecrest.Validate()
--- a/internal/remote/config/config.go
+++ b/internal/remote/config/config.go
@@ -18,6 +18,8 @@ limitations under the License.
 package config
 
 import (
+	"fmt"
+
 	amaltheadevv1alpha1 "github.com/SwissDataScienceCenter/amalthea/api/v1alpha1"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
@@ -35,8 +37,11 @@ const (
 )
 
 const (
-	serverPortFlag = "server-port"
-	fakeStartFlag  = "fake-start"
+	serverPortFlag         = "server-port"
+	fakeStartFlag          = "fake-start"
+	sessionPortFlag        = "session-port"
+	sessionURLPathFlag     = "session-url-path"
+	readinessProbeTypeFlag = "readiness-probe-type"
 )
 
 type RemoteSessionControllerConfig struct {
@@ -53,6 +58,15 @@ type RemoteSessionControllerConfig struct {
 
 	// FakeStart if true, do not start the remote session and print debug information
 	FakeStart bool
+
+	// SessionPort is the port where the remote session is expected to be serving
+	SessionPort int32
+
+	// SessionURLPath is the URL path for the HTTP readiness probe
+	SessionURLPath string
+
+	// ReadinessProbeType is "none", "tcp", or "http"
+	ReadinessProbeType string
 }
 
 func SetFlags(cmd *cobra.Command) error {
@@ -69,6 +83,30 @@ func SetFlags(cmd *cobra.Command) error {
 		return err
 	}
 	if err := viper.BindEnv(fakeStartFlag, configUtils.AsEnvVarFlag(fakeStartFlag)); err != nil {
+		return err
+	}
+
+	cmd.Flags().Int32(sessionPortFlag, 0, "port the remote session is expected to be serving on")
+	if err := viper.BindPFlag(sessionPortFlag, cmd.Flags().Lookup(sessionPortFlag)); err != nil {
+		return err
+	}
+	if err := viper.BindEnv(sessionPortFlag, configUtils.AsEnvVarFlag(sessionPortFlag)); err != nil {
+		return err
+	}
+
+	cmd.Flags().String(sessionURLPathFlag, "/", "URL path for the HTTP readiness probe")
+	if err := viper.BindPFlag(sessionURLPathFlag, cmd.Flags().Lookup(sessionURLPathFlag)); err != nil {
+		return err
+	}
+	if err := viper.BindEnv(sessionURLPathFlag, configUtils.AsEnvVarFlag(sessionURLPathFlag)); err != nil {
+		return err
+	}
+
+	cmd.Flags().String(readinessProbeTypeFlag, "", "readiness probe type: none, tcp, or http")
+	if err := viper.BindPFlag(readinessProbeTypeFlag, cmd.Flags().Lookup(readinessProbeTypeFlag)); err != nil {
+		return err
+	}
+	if err := viper.BindEnv(readinessProbeTypeFlag, configUtils.AsEnvVarFlag(readinessProbeTypeFlag)); err != nil {
 		return err
 	}
 
@@ -97,6 +135,9 @@ func GetConfig() (cfg RemoteSessionControllerConfig, err error) {
 
 	cfg.ServerPort = viper.GetInt32(serverPortFlag)
 	cfg.FakeStart = viper.GetBool(fakeStartFlag)
+	cfg.SessionPort = viper.GetInt32(sessionPortFlag)
+	cfg.SessionURLPath = viper.GetString(sessionURLPathFlag)
+	cfg.ReadinessProbeType = viper.GetString(readinessProbeTypeFlag)
 
 	return cfg, nil
 }
--- a/internal/remote/config/config.go
+++ b/internal/remote/config/config.go
@@ -102,7 +102,7 @@ func SetFlags(cmd *cobra.Command) error {
 		return err
 	}
 
-	cmd.Flags().String(readinessProbeTypeFlag, "", "readiness probe type: none, tcp, or http")
+	cmd.Flags().String(readinessProbeTypeFlag, "none", "readiness probe type: none, tcp, or http")
 	if err := viper.BindPFlag(readinessProbeTypeFlag, cmd.Flags().Lookup(readinessProbeTypeFlag)); err != nil {
 		return err
 	}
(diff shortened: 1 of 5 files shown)


internal/remote/server/server.go
```diff
@@ -114,7 +116,49 @@ func newServer(controller controller.RemoteSessionController) (server *echo.Echo
...
+				return c.NoContent(http.StatusOK)
+			}
+			conn, err := net.DialTimeout("tcp", fmt.Sprintf("127.0.0.1:%d", cfg.SessionPort), 5*time.Second)
+			if err != nil {
+				return c.NoContent(http.StatusServiceUnavailable)
+			}
+			if err := conn.Close(); err != nil {
+				slog.Error("failed to close readiness probe connection", "error", err)
+			}
+			return c.NoContent(http.StatusOK)
```
olevski, line 132: It is better to use the request context for this. That way if the request is cancelled the tcp connection is also cancelled. And we can still have the timeout we want.

```suggestion
		case string(amaltheadevv1alpha1.TCP):
			if cfg.SessionPort == 0 {
				return c.NoContent(http.StatusOK)
			}
            dialer := net.Dialer{}
			dialCtx, dialCtxCancel := context.WithTimeout(c.Request().Context(), 5*time.Second)
			defer dialCtxCancel()
			conn, err := dialer.DialContext(dialCtx, "tcp", fmt.Sprintf("127.0.0.1:%d", cfg.SessionPort))

			if err != nil {
				return c.NoContent(http.StatusServiceUnavailable)
			}
			if err := conn.Close(); err != nil {
				slog.Error("failed to close readiness probe connection", "error", err)
			}
			return c.NoContent(http.StatusOK)
```

internal/remote/server/server.go
```diff
@@ -114,7 +116,49 @@ func newServer(controller controller.RemoteSessionController) (server *echo.Echo
...
+			}
+			defer func() {
+				if err := resp.Body.Close(); err != nil {
+					slog.Error("failed to close readiness probe response body", "error", err)
+				}
+			}()
+			if resp.StatusCode >= 200 && resp.StatusCode < 400 {
+				return c.NoContent(http.StatusOK)
+			}
+			return c.NoContent(http.StatusServiceUnavailable)
```
olevski, line 157: Same as above, the request for the health check should be a child of the request to the server.

```suggestion
		case string(amaltheadevv1alpha1.HTTP):
			if cfg.SessionPort == 0 {
				return c.NoContent(http.StatusOK)
			}
			client := &http.Client{
				Timeout: 5 * time.Second,
				CheckRedirect: func(req *http.Request, via []*http.Request) error {
					return http.ErrUseLastResponse
				},
			}
			url := fmt.Sprintf("http://127.0.0.1:%d%s", cfg.SessionPort, cfg.SessionURLPath)
			req, err := http.NewRequestWithContext(c.Request().Context(), "GET", url, nil)
			if err != nil {
				return c.NoContent(http.StatusServiceUnavailable)
			}
			resp, err := client.Do(req)
			if err != nil {
				return c.NoContent(http.StatusServiceUnavailable)
			}
			defer func() {
				if err := resp.Body.Close(); err != nil {
					slog.Error("failed to close readiness probe response body", "error", err)
				}
			}()
			if resp.StatusCode >= 200 && resp.StatusCode < 400 {
				return c.NoContent(http.StatusOK)
			}
			return c.NoContent(http.StatusServiceUnavailable)
```

internal/remote/server/server_test.go
```diff
@@ -0,0 +1,206 @@
...
+			}(),
+			setupBackend: func() func() {
+				ln, err := net.Listen("tcp", "127.0.0.1:18000")
+				if err != nil {
+					panic(err)
+				}
+				return func() { _ = ln.Close() }
+			},
+			wantStatus: http.StatusOK,
+		},
```
olevski, line 83: This is a way to not hardcode the port to run the test server on.


```
func getFreePortOrDie() int {
	var a *net.TCPAddr
	var err error
	if a, err = net.ResolveTCPAddr("tcp", "localhost:0"); err == nil {
		var l *net.TCPListener
		if l, err = net.ListenTCP("tcp", a); err == nil {
			defer func() {
				err := l.Close()
				if err != nil {
					panic(err)
				}
			}()
			return l.Addr().(*net.TCPAddr).Port
		}
		panic(err)
	}
	panic(err)
}
```

From https://github.com/SwissDataScienceCenter/renku-frontend-buildpacks/blob/main/tests/e2e/utils.go#L162C1-L179C2

olevski (dismissed):
Thanks Salim. Just a few minor things. You can ignore the note about finding a free port if you wish.

internal/remote/config/config.go
```diff
@@ -72,6 +86,30 @@ func SetFlags(cmd *cobra.Command) error {
...
+
+	cmd.Flags().String(sessionURLPathFlag, "/", "URL path for the HTTP readiness probe")
+	if err := viper.BindPFlag(sessionURLPathFlag, cmd.Flags().Lookup(sessionURLPathFlag)); err != nil {
+		return err
+	}
+	if err := viper.BindEnv(sessionURLPathFlag, configUtils.AsEnvVarFlag(sessionURLPathFlag)); err != nil {
+		return err
+	}
+
+	cmd.Flags().String(readinessProbeTypeFlag, "", "readiness probe type: none, tcp, or http")
```
leafty: Why is the default `""` and not `"none"`?

internal/remote/server/server.go
```diff
@@ -114,7 +116,49 @@ func newServer(controller controller.RemoteSessionController) (server *echo.Echo
...
 	// Readiness endpoint
 	e.GET("/ready", func(c echo.Context) error {
-		return c.NoContent(http.StatusOK)
+		switch cfg.ReadinessProbeType {
+		case string(amaltheadevv1alpha1.None):
+			return c.NoContent(http.StatusOK)
+		case string(amaltheadevv1alpha1.TCP):
+			if cfg.SessionPort == 0 {
+				return c.NoContent(http.StatusOK)
+			}
```
leafty: This should not happen in my opinion: using a port value of `0` is invalid. The default values should map to `amaltheadevv1alpha1.None`.

internal/remote/server/server.go
```diff
@@ -114,7 +116,49 @@ func newServer(controller controller.RemoteSessionController) (server *echo.Echo
 
 	// Readiness endpoint
 	e.GET("/ready", func(c echo.Context) error {
-		return c.NoContent(http.StatusOK)
+		switch cfg.ReadinessProbeType {
+		case string(amaltheadevv1alpha1.None):
+			return c.NoContent(http.StatusOK)
```
leafty: Since we also have a `default` case, you can remove the `None` case.

internal/remote/config/config.go
```diff
@@ -97,11 +135,21 @@ func GetConfig() (cfg RemoteSessionControllerConfig, err error) {
...
 	return cfg, nil
 }
 
 func (cfg *RemoteSessionControllerConfig) Validate() error {
+	if cfg.ReadinessProbeType != "" &&
+		cfg.ReadinessProbeType != string(amaltheadevv1alpha1.None) &&
+		cfg.ReadinessProbeType != string(amaltheadevv1alpha1.TCP) &&
+		cfg.ReadinessProbeType != string(amaltheadevv1alpha1.HTTP) {
+		return fmt.Errorf("invalid readiness probe type: %s", cfg.ReadinessProbeType)
+	}
```
leafty, line 150: Is there any reason to not update the configuration to have `none` as the default? This way you can remove the defensive code in the endpoint implementation.

internal/remote/server/server_test.go
SalimKayal: thanks, added

internal/remote/config/config.go
SalimKayal: added default as none

internal/remote/config/config.go
SalimKayal: changed

internal/remote/server/server.go
SalimKayal: removed

internal/remote/server/server.go
SalimKayal: removed

internal/remote/server/server.go
```diff
@@ -114,7 +116,52 @@ func newServer(controller controller.RemoteSessionController) (server *echo.Echo
...
+				return c.NoContent(http.StatusServiceUnavailable)
+			}
+			if err := conn.Close(); err != nil {
+				slog.Error("failed to close readiness probe connection", "error", err)
+			}
+			return c.NoContent(http.StatusOK)
+		case string(amaltheadevv1alpha1.HTTP):
+			if cfg.SessionPort == 0 {
+				return c.NoContent(http.StatusOK)
+			}
```
leafty:
```suggestion
```

internal/remote/server/server.go
```diff
@@ -114,7 +116,52 @@ func newServer(controller controller.RemoteSessionController) (server *echo.Echo
...
+				if err := resp.Body.Close(); err != nil {
+					slog.Error("failed to close readiness probe response body", "error", err)
+				}
+			}()
+			if resp.StatusCode >= 200 && resp.StatusCode < 400 {
+				return c.NoContent(http.StatusOK)
+			}
+			return c.NoContent(http.StatusServiceUnavailable)
+		default:
+			// Unconfigured / unknown: preserve old behavior for backward compatibility
```
leafty:
```suggestion
			// Case None or unset: preserve old behavior for backward compatibility
```

leafty (approved):

PR #1136, opened 2026-06-15
feat: fetch remote FirecREST logs into sessions

SalimKayal:
## Summary

Fetch remote Slurm session stdout/stderr logs via FirecREST and stream them into the controller pod's stdout, making them visible through `kubectl logs`.

## Motivations and Context

Remote sessions running on Slurm clusters write their output to files on the cluster filesystem (e.g., `slurm-<jobid>.out`). Previously, these logs were inaccessible from the Kubernetes sidecar, making it hard for users and operators to debug running or completed remote sessions. This change polls the log files through FirecREST's filesystem view API and forwards any new lines to the pod's stdout.

## Changes

- **Added `fetchSessionLogs`** — polls remote stdout and stderr files via `GetViewFilesystemSystemNameOpsViewGet` and writes complete lines to `os.Stdout` prefixed with `[session/stdout]` or `[session/stderr]`.
- **Added `fetchLogStream`** — handles a single stream: buffered reading, byte-offset tracking, and line-by-line flushing so partial lines survive across polls.
- **Extended `savedState`** — now persists `stdout_path`, `stderr_path`, `stdout_offset`, and `stderr_offset` alongside the existing `job_id`.
- **Renamed `saveJobID` → `saveState` and `recoverJobID` → `recoverState`** — updated to serialize and deserialize the new fields.
- **Updated `periodicSessionStatus`** — calls `fetchSessionLogs` on each status tick and persists offsets to disk.
- **Updated `Start`** — constructs expected Slurm log file paths after job submission and calls `saveState`.

## Notes

- Log fetch errors are logged at `Warn` level and do not fail the status loop; missing files early in the job lifecycle are silently ignored.
- The buffer/offset approach ensures we don't duplicate lines across polls or after a controller restart.

### DIFF
--- a/internal/remote/firecrest/controller.go
+++ b/internal/remote/firecrest/controller.go
@@ -651,11 +651,29 @@ func (c *FirecrestRemoteSessionController) fetchSessionLogs(ctx context.Context)
 		return
 	}
 
-	c.fetchLogStream(ctx, "stdout", c.stdoutPath, &c.stdoutOffset, &c.stdoutBuf)
-	c.fetchLogStream(ctx, "stderr", c.stderrPath, &c.stderrOffset, &c.stderrBuf)
+	c.fetchLogStream(ctx, "stdout")
+	c.fetchLogStream(ctx, "stderr")
 }
 
-func (c *FirecrestRemoteSessionController) fetchLogStream(ctx context.Context, stream string, filePath string, offset *int, buf *bytes.Buffer) {
+func (c *FirecrestRemoteSessionController) fetchLogStream(ctx context.Context, stream string) {
+	var filePath string
+	var offset *int
+	var buf *bytes.Buffer
+
+	switch stream {
+	case "stdout":
+		filePath = c.stdoutPath
+		offset = &c.stdoutOffset
+		buf = &c.stdoutBuf
+	case "stderr":
+		filePath = c.stderrPath
+		offset = &c.stderrOffset
+		buf = &c.stderrBuf
+	default:
+		slog.Warn("unknown stream type", "stream", stream)
+		return
+	}
+
 	size := 524288 // 512KB
 	params := GetViewFilesystemSystemNameOpsViewGetParams{
 		Path:   filePath,
--- a/internal/remote/firecrest/controller.go
+++ b/internal/remote/firecrest/controller.go
@@ -361,6 +361,9 @@ func (c *FirecrestRemoteSessionController) Start(ctx context.Context) error {
 				p = path.Join(sessionPath, p)
 			}
 			c.stderrPath = p
+		} else {
+			// Slurm writes stderr to stdout when StandardError is not explicitly set.
+			c.stderrPath = c.stdoutPath
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


leafty (commented):
Why read from the file system when stdout and stderr are returned from [`GET /compute/{system_name}/jobs/{job_id}/metadata`](https://eth-cscs.github.io/firecrest-v2/openapi/#/compute/get_job_metadata_compute__system_name__jobs__job_id__metadata_get)?

> Get metadata of a job by {job_id}

internal/remote/firecrest/controller.go
```diff
@@ -599,6 +621,61 @@ func (c *FirecrestRemoteSessionController) getCurrentStatus(ctx context.Context)
...
+		return
+	}
+
+	content := *res.JSON200.Output
+	if content == "" {
+		return
+	}
+
+	*offset += len(content)
+
```
olevski: question: Why not get this method to just modify `c.stdoutOffset` and the buffer rather than passing it. It already has a pointer to the session controller and access to both. You can have a switch statement inside to pick the right offset, buffer and file patch based on whether the stream is `stdout` or `stderr`.

internal/remote/firecrest/controller.go
```diff
@@ -599,6 +621,61 @@ func (c *FirecrestRemoteSessionController) getCurrentStatus(ctx context.Context)
...
+		if idx == -1 {
+			break
+		}
+		line := string(data[:idx])
+		if _, err := fmt.Fprintf(os.Stdout, "[session/%s] %s\n", stream, line); err != nil {
+			slog.Warn("failed to write session log", "stream", stream, "error", err)
+		}
+		buf.Next(idx + 1)
+	}
+}
```
olevski, line 730: I think we have a problem here if I am not mistaken. The offset is based on all the data read in the buffer. But you print to stdout stuff that is only in a whole line. So if you have a partial line then the offset is updated including the stuff in the partial line. But the partial line is never printed. 

So if you received something like this you would have stuff missing.


Tick 0:
```
line 1\n
line 2\n
li
```

Tick 1:
```
line 1\n
line 2\n
line 3\n
```

I think what will go to stdout will be:
```
line 1
line 2
ne 3
```

Because at tick 1 the offset will include the `li` partial from the last line.

olevski (changes requested):

SalimKayal:
> Why read from the file system when stdout and stderr are returned from [`GET /compute/{system_name}/jobs/{job_id}/metadata`](https://eth-cscs.github.io/firecrest-v2/openapi/#/compute/get_job_metadata_compute__system_name__jobs__job_id__metadata_get)?

thanks I missed this one.

leafty:
> > Why read from the file system when stdout and stderr are returned from [`GET /compute/{system_name}/jobs/{job_id}/metadata`](https://eth-cscs.github.io/firecrest-v2/openapi/#/compute/get_job_metadata_compute__system_name__jobs__job_id__metadata_get)?
> 
> thanks I missed this one.

There may be a good case for reading from the file system, just making sure we are considering the simplest option first.

internal/remote/firecrest/controller.go
SalimKayal: fixed offset for partial lines in 619df23

SalimKayal:
I read the path for the stdout/err files from the metadata with a fallback for slurm defaults in 594448c.

internal/remote/firecrest/controller.go
SalimKayal: refactored to use the pointers in 18f0a45.

SalimKayal:
FYI this is what FirecREST runs to get the metadata. https://github.com/eth-cscs/firecrest-v2/blob/master/src/lib/scheduler_clients/slurm/cli_commands/sacct_job_metadata_command.py#L12

olevski (dismissed):

olevski (approved):

PR #1140, opened 2026-06-17
chore: Simplify version checking logic in install_rclone

sambuc:
### PR Stack
* #1137
* #1138
* #1139
* 🐸 #1140
* #1141
* #1142
* #1143
* #1144
* #1145
* #1146
* #1147

### DIFF
--- a/internal/remote/firecrest/session_script.sh
+++ b/internal/remote/firecrest/session_script.sh
@@ -40,21 +40,16 @@ function install_rclone() {
     rclone_pkg="${RENKU_PKG}/rclone/v${rclone_version}"
     rclone_bin="${rclone_pkg}/rclone"
 
-    skip_install="0"
     if [ -f "${rclone_bin}" ]; then
         version="$("${rclone_bin}" version || echo "bad executable")"
         version="$(echo "${version}" | head -n 1)"
         expected="rclone v${rclone_version}"
         if [ "${version}" = "${expected}" ]; then
-            skip_install="1"
+            echo "${rclone_bin}"
+            return 0
         else
             >&2 echo "WARNING: found mismatching rclone version ${version}"
         fi
-    fi
-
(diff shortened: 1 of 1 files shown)


internal/remote/firecrest/session_script.sh
```diff
@@ -40,23 +40,18 @@ function install_rclone() {
...
     rclone_bin="${rclone_pkg}/rclone"
 
-    skip_install="0"
     if [ -f "${rclone_bin}" ]; then
         version="$("${rclone_bin}" version || echo "bad executable")"
         version="$(echo "${version}" | head -n 1)"
         expected="rclone v${rclone_version}"
-        if [ "${version}" = "${expected}" ]; then
-            skip_install="1"
+        if [ "x${version}" = "x${expected}" ]; then
```
leafty: why `"x${version}"`?

sambuc: This is something I have learned to do to protect against empty strings, which would yield a syntax error, if one of the side returns empty.

I did this by habit, but it is not useful here, as we guarantee `rclone v` here, in the specific case.

sambuc: Although, `${version}` could be empty in case the command returns true, but outputs nothing.

leafty: The empty string WILL NOT yield a syntax error.

leafty: This bash, not some other shell.

sambuc: I did a few test, this can happen, but only when you do not quote your two values, in that case you get it. 

I guess I picked up the habit from someone who did not realise that when double-quoting your left value, then you did not need this anymore.

I'll patch this to remove the `x`es.

leafty (approved):

PR #1148, opened 2026-06-18
chore: Change environment into a nullable param

sambuc:
### PR Stack
* 🐸 #1148
* #1149
* #1150
* #1151
* #1152

### DIFF
--- a/internal/remote/firecrest/session_script.sh
+++ b/internal/remote/firecrest/session_script.sh
@@ -140,7 +140,7 @@ if !(nvidia-smi 2>&1 >/dev/null); then
     export NVIDIA_VISIBLE_DEVICES=void
 fi
 
-if srun --help |grep -q -- --environment; then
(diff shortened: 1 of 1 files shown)


internal/remote/firecrest/session_script.sh
```diff
@@ -140,12 +140,17 @@ if !(nvidia-smi 2>&1 >/dev/null); then
     export NVIDIA_VISIBLE_DEVICES=void
 fi
 
-# Create the environment.toml file to run the session
-EDF_FILE="${SESSION_DIR}/environment.toml"
-cat <<EOF >"${EDF_FILE}"
+if srun --help |grep -q -- --environment; then
```
leafty: nitpick


```suggestion
if srun --help | grep -q -- --environment; then
```

leafty (approved):

PR #1149, opened 2026-06-18
chore: Improve script cleanup call

sambuc:
### PR Stack
* #1148
* 🐸 #1149
* #1150
* #1151
* #1152

### DIFF
--- a/internal/remote/firecrest/session_script.sh
+++ b/internal/remote/firecrest/session_script.sh
@@ -243,15 +243,19 @@ if [ -n "${GIT_REPOSITORIES}" ]; then
     fi
 fi
 
-exit_script() {
+function exit_script() {
     echo "Cleaning up session..."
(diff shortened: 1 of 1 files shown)


internal/remote/firecrest/session_script.sh
```diff
@@ -243,15 +243,19 @@ if [ -n "${GIT_REPOSITORIES}" ]; then
     fi
 fi
 
-exit_script() {
+function exit_script() {
     echo "Cleaning up session..."
-    # fusermount3 -u "${SESSION_WORK_DIR}/era5" || true
```
leafty, line 248: Note: I think you should keep the unmount placeholder.

sambuc: I'll keep it until the code to actually unmount things arrives.

sambuc: (I mean I have put back the line as you suggested)

leafty (approved):

PR #1151, opened 2026-06-18
chore: Reorganise session folder for hpc sessions

sambuc:
Introduce two sub folders to separate the session secrets from the data
connector configurations as well as the infra secrets.

### PR Stack
* #1148
* #1149
* #1150
* 🐸 #1151
* #1152

### DIFF
--- a/internal/remote/firecrest/session_script.sh
+++ b/internal/remote/firecrest/session_script.sh
@@ -37,6 +37,8 @@ esac
 : ${SESSION_DIR:="${PWD}"}
 : ${SESSION_WORK_DIR:="${SESSION_DIR}/work"}
 : ${SECRETS_DIR:="${SESSION_DIR}/secrets"}
+: ${SECRETS_SLOTS_DIR:="${SECRETS_DIR}/slots/"}
+: ${SECRETS_DCS_DIR:="${SECRETS_DIR}/dcs"}
 : ${LOGS_DIR:="${SESSION_DIR}/logs"}
 
 # Setup session environment
@@ -129,7 +131,10 @@ function install_wstunnel() {
     echo "${wstunnel_bin}"
 }
 
-for d in SESSION_WORK_DIR SECRETS_DIR LOGS_DIR
+for d in \
+    SESSION_DIR \
+    SESSION_WORK_DIR \
+    LOGS_DIR
 do
     echo "${d}: ${!d}"
     mkdir -p "${!d}"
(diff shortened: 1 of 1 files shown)


internal/remote/firecrest/session_script.sh
```diff
@@ -37,6 +37,8 @@ esac
 : ${SESSION_DIR:="${PWD}"}
 : ${SESSION_WORK_DIR:="${SESSION_DIR}/work"}
 : ${SECRETS_DIR:="${SESSION_DIR}/secrets"}
+: ${SECRETS_SLOTS_DIR:="${SECRETS_DIR}/slots/"}
```
leafty: Let's call them "user secrets":


```suggestion
: ${SECRETS_SLOTS_DIR:="${SECRETS_DIR}/user/"}
```

internal/remote/firecrest/session_script.sh
```diff
@@ -37,6 +37,8 @@ esac
...
+: ${SECRETS_DCS_DIR:="${SECRETS_DIR}/dcs"}
```
leafty: Explicit naming:


```suggestion
: ${SECRETS_DATA_CONNECTORS_DIR:="${SECRETS_DIR}/data_connectors"}
```

internal/remote/firecrest/session_script.sh
sambuc: ok, applied

internal/remote/firecrest/session_script.sh
sambuc: Applied as well

internal/remote/firecrest/session_script.sh
```diff
@@ -129,7 +131,10 @@ function install_wstunnel() {
     echo "${wstunnel_bin}"
 }
 
-for d in SESSION_WORK_DIR SECRETS_DIR LOGS_DIR
+for d in \
+    SESSION_DIR \
+    SESSION_WORK_DIR \
+    LOGS_DIR
```
leafty, line 140: I think you need to have:
- `SESSION_WORK_DIR` (with `-p` means we can skip `SESSION_DIR`)
- `SECRETS_USER_DIR`
- `SECRETS_DATA_CONNECTORS_DIR`
- `LOGS_DIR`

leafty (changes requested):

internal/remote/firecrest/session_script.sh
sambuc: Thank you for the review, I have added the missing folder variables.

I am explicitly listing all the folder variables in the loop to avoid assumptions about their relationships.

Calling `mkdir` on a folder which already exists is a no-op, so this should not cause any problems.

olevski (approved):

leafty (approved):

PR #1153, opened 2026-06-19
fix: Return correct non-interactive session state when pod is gone 

eikek:
It happens that the pod is gone, but the job and amalthea session still exists. In this case the status of the amalthea session needs to be derived from the job.

/deploy

### DIFF
--- a/internal/controller/children.go
+++ b/internal/controller/children.go
@@ -547,7 +547,7 @@ func (c ChildResourceUpdates) failureMessage(pod *v1.Pod, job *batchv1.Job) stri
 	if msg != "" {
 		return msg
 	}
-	msg = jobFailureReason(c.Job.Manifest)
+	msg = jobFailureReason(job)
 	if msg != "" {
 		return msg
 	}
--- a/internal/controller/children.go
+++ b/internal/controller/children.go
@@ -520,7 +520,7 @@ func (c ChildResourceUpdates) State(cr *amaltheadevv1alpha1.AmaltheaSession, pod
 	case jobIsSuccess(job):
 		return amaltheadevv1alpha1.Succeeded, ""
 	case jobIsFailed(job):
-		return amaltheadevv1alpha1.Failed, "The job failed."
+		return amaltheadevv1alpha1.Failed, "The job has finished with a failure status."
 	default:
 		return amaltheadevv1alpha1.NotReady, ""
 	}
--- a/internal/controller/children.go
+++ b/internal/controller/children.go
@@ -503,7 +503,7 @@ func (c ChildResourceUpdates) IsRunning(pod *v1.Pod) bool {
 }
 
 func (c ChildResourceUpdates) State(cr *amaltheadevv1alpha1.AmaltheaSession, pod *v1.Pod, job *batchv1.Job) (amaltheadevv1alpha1.State, string) {
-	msg := c.failureMessage(pod)
+	msg := c.failureMessage(pod, job)
 	switch {
 	case cr.GetDeletionTimestamp() != nil:
 		return amaltheadevv1alpha1.NotReady, ""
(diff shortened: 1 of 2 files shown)


internal/controller/children.go
```diff
@@ -517,6 +517,10 @@ func (c ChildResourceUpdates) State(cr *amaltheadevv1alpha1.AmaltheaSession, pod
 		return amaltheadevv1alpha1.Running, ""
 	case podIsCompleted(pod):
 		return amaltheadevv1alpha1.Succeeded, ""
+	case jobIsSuccess(job):
+		return amaltheadevv1alpha1.Succeeded, ""
```
sgaist:
```suggestion
	case podIsCompleted(pod):
		fallthrough
	case jobIsSuccess(job):
		return amaltheadevv1alpha1.Succeeded, ""
```

internal/controller/children.go
```diff
@@ -748,10 +752,17 @@ func (c ChildResourceUpdates) Status(
 			log.Error(err, "Could not read the session pod when updating the status")
 		}
 	}
+	job, err := cr.GetJob(ctx, r.Client)
+	if err != nil {
+		job = nil
```
sgaist, line 755: I think you don't need to do this as `job` should already be `nil` based on the implementation of `GetJob`
```suggestion
```

sgaist (changes requested):
Some small things otherwise looks good

internal/controller/children.go
eikek: Yes, but the impl might change and not immediatly noticable to all usages. Judging from the signature, it might not be nil 🤷🏼. The code above for pods looks the same. I would rather keep this line, unless there are strong opinions for removal.

sgaist: I missed the one from pods but I would have said the same.

In any case, I won't block on that but if there's this kind of doubt, I think we should have unit tests in place that screams at us when function APIs change in such a significant way.

eikek: There should be this doubt all the time with `nil` sitting around the corner :-) I don't really understand the strong opinion here, tbh. It's not wrong to explicitly set `job` to `nil` on error, because the function doesn't return early as it is often the case. So here it can make sure to not accidentally use a "bad" job object - like an empty one, which is also easily done in Go. It might not be necessary in all cases at runtime, but I don't see anything flat wrong here. 🤷🏼

sgaist (dismissed):

internal/controller/status.go
```diff
@@ -81,6 +81,30 @@ func podIsReady(pod *v1.Pod) bool {
...
+			return true
+		}
+	}
+	return false
+}
+
+func jobIsFailed(job *batchv1.Job) bool {
+	if job == nil {
+		return false
+	}
```
leafty: Shouldn't this branch return true?

internal/controller/children.go
```diff
@@ -516,7 +516,11 @@ func (c ChildResourceUpdates) State(cr *amaltheadevv1alpha1.AmaltheaSession, pod
 	case c.IsRunning(pod):
 		return amaltheadevv1alpha1.Running, ""
 	case podIsCompleted(pod):
+		fallthrough
+	case jobIsSuccess(job):
 		return amaltheadevv1alpha1.Succeeded, ""
+	case jobIsFailed(job):
+		return amaltheadevv1alpha1.Failed, "The job has finished with a failure status."
```
leafty: Would it be possible to forward the error from the job? It may contain the non-zero exit code or other message that could be useful to explain the failure.

internal/controller/status.go
eikek: Not sure - if the job is `nil` there is no way of telling whether failed or not, right? Would you rather `true`? Also for `jobIsSuccess`?

internal/controller/children.go
eikek: Yes, good point! I try to return the message if it is possible.

sgaist (dismissed):

leafty (dismissed):

eikek:
I tested it on a [data services pr](https://github.com/SwissDataScienceCenter/renku-data-services/pull/1361). The deployment [was successful](https://github.com/SwissDataScienceCenter/amalthea/actions/runs/28112551464) when using the `what-da-heck` branch of renku-actions

leafty (approved):

PR #1155, opened 2026-06-24
feat: define a run ID per continuous execution of a session

leafty:
Define `runID` which is identifies a continuous execution of a session; in other words, a new `runID` is assigned every time a session is resumed.

The `runID` is propagated to the AmaltheaSession's status field, to its annotations and to the child resources annotations.

The new pod annotations (`renku.io/run_id` and `renku.io/session_uid`) will allow logs to be attributed to a specific "run" so that old logs from a previous run can be discarded earlier than the logs for a current or recent run.

/deploy


api/v1alpha1/amaltheasession_children.go
```diff
@@ -218,12 +218,26 @@ func (cr *AmaltheaSession) StatefulSet(cfg config.AmaltheaSessionConfiguration)
...
+		annotations[key] = cr.Spec.Template.Metadata.Annotations[key]
+	}
+	// Add the annotations for the runID and the launchID
+	if cr.Status.RunID != "" {
+		annotations["renku.io/run_id"] = cr.Status.RunID
+	}
+	uid := string(cr.GetUID())
+	if uid != "" {
+		annotations["renku.io/launch_id"] = uid
+	}
```
olevski, line 233: question: Why not call it `renku.io/k8s_uid`? Then it is clear what it is. This way someone has to tell you it is the same as the uid.

And why do we need this in the annotations in the first place where you can just get it from the resource metadata?

olevski (changes requested):

api/v1alpha1/amaltheasession_children.go
leafty: I can name it `renku.io/session_uid`.

> And why do we need this in the annotations in the first place where you can just get it from the resource metadata?

The goal is to have metadata in Loki so that log lines can be attributed to sessions. The metadata available can only come from the pod, not from a parent resource which is why I propagate the metadata down to pod labels and annotations.

See: https://github.com/SwissDataScienceCenter/terraform-renku/pull/3201

olevski (approved):

Issue #1154, opened 2026-06-15
When there is autoscaling that  takes a while (more than a minute or so) the session shows up as failed


olevski:
When the session gets scheduled the failed status clears. But it is not clear what is going on to users.

<img width="1494" height="1216" alt="Image" src="https://github.com/user-attachments/assets/702ab7a1-f001-47cc-92ef-b4d9052d074a" />

<img width="1677" height="534" alt="Image" src="https://github.com/user-attachments/assets/aa60c3aa-c1f9-4035-924f-9356579615a1" />

Seen in data svc 0.76.0.

I think in amalthea we have a  timeout for how long we wait for the session to be scheduled before we mark it as failed. This may have to be adjusted. I also thought that Amalthea tries to catch the pod events that indicate that autoscaling was started. But it does not or the event names have changed.

leafty:
Wrong logic in amalthea: https://github.com/SwissDataScienceCenter/amalthea/blob/1488d77ab7adb5a7507d52beb376414061041e4d/internal/controller/children.go#L579

`Scheduled` or `TriggeredScaleUp` events will not appear in any specific order compared to `FailedScheduling` events. If any of `Scheduled` or `TriggeredScaleUp` happens, the pod should get scheduled soon.

leafty:
@eikek see above: events do not have a specific ordering, which means the current logic is incorrect.

eikek:
@leafty What about a case, when there is a `TriggeredScaleUp` but then scaling fails due to resource issues (or whatever)?

leafty:
> [@leafty](https://github.com/leafty) What about a case, when there is a `TriggeredScaleUp` but then scaling fails due to resource issues (or whatever)?

I don't know exactly what are the timeout we have, but we can try to mark failed if the last relevant scheduling event is too old.

leafty:
@eikek I think you may be confused about how k8s pod scheduling and node autoscaling works.

Both pod scheduling and node auto-scaling are handled by controllers which run their own reconcile loop (similar to amalthea).
In the logs posted above, this is what is happening (simplified):
1. The pod is created (from the StatefulSet)
2. The pod creation events triggers the reconcile loop of `default-scheduler` (the default pod scheduler -- note that specific k8s infra like knative may use another pod scheduler)
    a. The scheduler sees that no eligible node can take on the pod workload => it emits a `FailedScheduling` event
[...]

PR #1157, opened 2026-06-26
Wip: try to improve detecting autoscale failure

eikek:
If it is (somewhat reliably) determined that an existing autoscale failed, return early failed condition. otherwise, the timeout is increased until a final failure is returned

Refs: #1154 

/deploy

### DIFF
--- a/.github/workflows/deploy-pr.yaml
+++ b/.github/workflows/deploy-pr.yaml
@@ -97,7 +97,7 @@ jobs:
           echo "KUBECONFIG=${KUBECONFIG}" >> "$GITHUB_ENV"
 
       - name: Build and deploy
-        uses: SwissDataScienceCenter/renku-actions/deploy-renku@v1.22.1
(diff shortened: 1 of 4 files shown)


leafty (commented):
Why do we introduce more complex logic here? I think what needs to be fixed is the current lookup of conditions and properly take into account the TriggeredScaleUp event.

eikek:
> Why do we introduce more complex logic here? I think what needs to be fixed is the current lookup of conditions and properly take into account the TriggeredScaleUp event.

I'm sorry, I went down the wrong path! I talked to @olevski and now the timeout is increased when we determine a failed session start. The reason is that `TriggerScaleUp` could still allow for the auto-scale to fail eventually. It seems too hard/not possible to distinguish this from the pod events.

leafty (approved):
OK 👍 

Maybe we could find a way to propagate the scale up event to the user so that we could display "A new node is starting, this may take a while..."

eikek:
Oh no, I'm sorry - i didn't change the PR title before merge :-(

PR #1159, opened 2026-07-08
feat: Upload secrets to HPC cluster

sambuc:
### PR Stack
* 🐸 #1159
* #1160
* #1161
* #1162
* #1163

### DIFF
--- a/internal/remote/firecrest/controller.go
+++ b/internal/remote/firecrest/controller.go
@@ -126,6 +126,58 @@ func (c *FirecrestRemoteSessionController) Status(ctx context.Context) (state mo
 	return c.currentStatus, c.currentStatusError
 }
 
+func (c *FirecrestRemoteSessionController) uploadSecrets(ctx context.Context, remoteSecretsPath string) error {
+	var err error
+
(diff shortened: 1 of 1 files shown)


internal/remote/firecrest/controller.go
```diff
@@ -211,8 +263,17 @@ func (c *FirecrestRemoteSessionController) Start(ctx context.Context) error {
 		if err != nil {
 			return err
 		}
+		err = c.chmod(startCtx, path.Join(secretsPath, "wstunnel_secret"), "400")
+		if err != nil {
+			return err
+		}
+	}
+	// Upload user secrets into secretsPath
+	secretSlotsPath := path.Join(secretsPath, "slots")
```
leafty: These are not called slots. The slot is the "shape" of the secret, it tells the user what type of secret they need to provide and which filename it will populate.

```suggestion
	secretSlotsPath := path.Join(secretsPath, "user")
```

sambuc: Sorry, this went past me, we had discussed this already. I'll make sure this is fixed everywhere it needs to be.

leafty (approved):

PR #1161, opened 2026-07-08
chore: Add OIDC config only when needed

sambuc:
### PR Stack
* #1159
* #1160
* 🐸 #1161
* #1162
* #1163

### DIFF
--- a/api/v1alpha1/amaltheasession_children.go
+++ b/api/v1alpha1/amaltheasession_children.go
@@ -779,7 +779,7 @@ func (as *AmaltheaSession) Secret() v1.Secret {
 	}
 
 	// Skip the 'oidc' configuration if it is not needed
-	if as.Spec.Authentication.Type == Oidc {
+	if as.Spec.Authentication != nil && as.Spec.Authentication.Type == Oidc {
 		pathPrefix := as.ingressPathPrefix()
 		sessionURL := as.GetURL()
 		pathPrefixURL := url.URL{Host: sessionURL.Host, Path: pathPrefix, Scheme: sessionURL.Scheme}
--- a/api/v1alpha1/amaltheasession_children.go
+++ b/api/v1alpha1/amaltheasession_children.go
@@ -778,7 +778,7 @@ func (as *AmaltheaSession) Secret() v1.Secret {
 		return secret
 	}
 
-	// Skip the 'oidc' configuration if it is not needed
+	// Add the 'oidc' configuration if requested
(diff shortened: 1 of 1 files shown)


api/v1alpha1/amaltheasession_children.go
```diff
@@ -778,66 +778,69 @@ func (as *AmaltheaSession) Secret() v1.Secret {
...
-		"upstreams": []map[string]any{
-			{
-				"id":                    "amalthea-upstream",
-				"path":                  pathPrefix,
-				"uri":                   fmt.Sprintf("http://127.0.0.1:%d", upstreamPort),
-				"insecureSkipTLSVerify": true,
-				"passHostHeader":        true,
-				"proxyWebSockets":       true,
+	// Skip the 'oidc' configuration if it is not needed
+	if as.Spec.Authentication != nil && as.Spec.Authentication.Type == Oidc {
```
leafty: Why this change? From line 772: `as.Spec.Authentication.Type` must equal `Oidc` so this new `if` branch does nothing.

sambuc: This is to prepare the next patch, as the code paths are a bit intertwined.

By splitting this in two PR, it makes it a lot simpler to understand what is going on.

Here mostly adding an if and some indenting, and in #1162 taking care of the tunnel secret.

leafty: I looked at that and this comment still applies.

sambuc: In the following patch, line 772 is removed, so I am confused by your reply.

api/v1alpha1/amaltheasession_children.go
```diff
@@ -778,66 +778,69 @@ func (as *AmaltheaSession) Secret() v1.Secret {
...
-	upstreamConfig := map[string]any{
-		"upstreams": []map[string]any{
-			{
-				"id":                    "amalthea-upstream",
-				"path":                  pathPrefix,
-				"uri":                   fmt.Sprintf("http://127.0.0.1:%d", upstreamPort),
-				"insecureSkipTLSVerify": true,
-				"passHostHeader":        true,
-				"proxyWebSockets":       true,
+	// Skip the 'oidc' configuration if it is not needed
```
leafty: You need to change this comment as it is confusing.


```suggestion
	// Add the 'oidc' configuration if requested
```

leafty (approved):

PR #1162, opened 2026-07-08
chore: Refactor tunnelSecret handling

sambuc:
### PR Stack
* #1159
* #1160
* #1161
* 🐸 #1162
* #1163

PR #1163, opened 2026-07-08
chore: Rename path variables for clarity

sambuc:
### PR Stack
* #1159
* #1160
* #1161
* #1162
* 🐸 #1163

### DIFF
--- a/internal/remote/firecrest/controller.go
+++ b/internal/remote/firecrest/controller.go
@@ -182,6 +182,10 @@ func (c *FirecrestRemoteSessionController) uploadSecrets(ctx context.Context, re
 //
 //nolint:gocyclo // TODO: can we break down session start?
 func (c *FirecrestRemoteSessionController) Start(ctx context.Context) error {
+	// Please note:
+	// * local*Path: A path on the current system
+	// * remote*Path: A path on the remote firecrest system
+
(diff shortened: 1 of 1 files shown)


leafty (commented):
Not sure I understand which type of clarity I gain by having `srun` prefixed everywhere.

sambuc:
I spend quite some time before realising this code was managing 3 sets of paths:
 1. The paths in the proxy container
 2. The paths on the node as seen by the srun command.
 3. The paths within the (final) user container

And figuring out which was which and when it was used.

Maybe I should have prefixed with something else but it was necessary for me to be a bit more explicit about what the variables were holding.

leafty:
> I spend quite some time before realising this code was managing 3 sets of paths:
> 
> 1. The paths in the proxy container
> 2. The paths on the node as seen by the srun command.
> 3. The paths within the (final) user container
> 
> And figuring out which was which and when it was used.
> 
> Maybe I should have prefixed with something else but it was necessary for me to be a bit more explicit about what the variables were holding.

How about:
* `localX` local file path, can be read from `os.Read()`
* `remoteX`, remote file path, created by calling Firecrest
* `containerX`, path inside the session container

olevski (approved):

PR #1165, opened 2026-07-09
fix: set working directory in the session

olevski:


leafty (commented):
The change looks OK, though I am not sure what we are gaining from this.

The guidance for external images is always to rely on the env variables `RENKU_MOUNT_DIR` and `RENKU_WORKING_DIR` to know where the Renku files and folders are located.

leafty (approved):

eikek (approved):

PR #1167, opened 2026-07-09
fix: clear failed scheduling since status

### LINKED ISSUE #1166: FailedSchedulingSince is not cleared after a session gets to run
The `status.failedSchedulingSince` field is not cleared once the session is Running (or is scheduled) which means that pausing then resuming may show the session as immediately failing. The `status.failedSchedulingSince` field should hold no value when any of these conditions is true:
* Session is Running
* Session is paused (`hibernated: true`)
* Pod is scheduled (this may not be trivial to detect, so the above two conditions should fix the bug)

olevski:
closes https://github.com/SwissDataScienceCenter/amalthea/issues/1166

/deploy

### DIFF
--- a/internal/controller/children.go
+++ b/internal/controller/children.go
@@ -877,6 +877,10 @@ func (c ChildResourceUpdates) Status(
 		status.InitContainerCounts.Ready = 0
 	}
 
+	if state == amaltheadevv1alpha1.Failed || state == amaltheadevv1alpha1.Hibernated || state == amaltheadevv1alpha1.Running || state == amaltheadevv1alpha1.Succeeded {
+		status.FailedSchedulingSince = metav1.Time{}
+	}
+
 	c.statusCallback(&status)
 
 	// Used for debugging to ensure the reconcile loop does not needlessly reschdule or update child resources
--- a/internal/controller/children.go
+++ b/internal/controller/children.go
@@ -123,6 +123,7 @@ func (c ChildResource[T]) Reconcile(ctx context.Context, clnt client.Client, cr 
 				statusCallback = func(status *amaltheadevv1alpha1.AmaltheaSessionStatus) {
 					status.IdleSince = metav1.Time{}
 					status.FailingSince = metav1.Time{}
+					status.FailedSchedulingSince = metav1.Time{}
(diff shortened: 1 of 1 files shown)


olevski:
@eikek was there a reason why you did not try to find the `scheduled` event in the events? Is it unreliable?

olevski:
Note: I may have to push one more commit to cleanup some of the tests.

internal/controller/children.go
```diff
@@ -593,19 +599,33 @@ func EventsInferedState(ctx context.Context, cr *amaltheadevv1alpha1.AmaltheaSes
...
+				log.Info("Found FailedScheduling event, finally failing", "waited", waitedTime)
+				return EisrFinallyFailed, fmt.Errorf("failed scheduling: %s", failedSchedulingEvent.Message)
 			} else {
-				if waitedTime >= maxWaitForClearFailedScheduling {
-					log.Info("Found FailedScheduling event, finally failing", "waited", waitedTime)
-					return EisrFinallyFailed, fmt.Errorf("failed scheduling: %s", v.Message)
-				} else {
-					log.Info("Found FailedScheduling event, temporary failing", "event", v.Message, "waiting", waitedTime)
-					return EisrTemporaryFailed, nil
-				}
```
olevski, line 608: All of this has moved into the `if failedSchedulingEvent != nil {` clause in the new code.

olevski: It will be clearer what I am talking about if you look at the side-by-side diff view of the PR and see what this comment points to.

leafty (approved):
LGTM

eikek:
> @eikek was there a reason why you did not try to find the `scheduled` event in the events? Is it unreliable?

I can't quite remember. The whole method of looking into the pod events is only called when the current state is "NotReady". So maybe I didn't even consider a scheduled event, because the logic of resetting this timestamp is/should be somewhere else. I also can't remember messing with the `FailedSchedulingSince` time in the last change 🤔 . So I'm really wondering why such a big change is necessary, but perhaps I removed accidentally some lines in the list change 🙈

eikek (approved):

PR #1173, opened 2026-07-16
feat: Upload DCs secrets

sambuc:
### PR Stack
* 🐸 #1173
* #1174

### DIFF
--- a/api/v1alpha1/amaltheasession_children.go
+++ b/api/v1alpha1/amaltheasession_children.go
@@ -4,16 +4,19 @@ import (
 	"context"
 	"crypto/rand"
 	"encoding/base64"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
 	"maps"
 	"net/url"
 	"os"
+	"path"
 	"sort"
 	"strings"
 	"time"
 
+	"github.com/SwissDataScienceCenter/amalthea/internal/common"
 	"github.com/SwissDataScienceCenter/amalthea/internal/controller/config"
 	"gopkg.in/yaml.v3"
 	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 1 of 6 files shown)


api/v1alpha1/amaltheasession_children.go
```diff
@@ -642,11 +645,92 @@ func (cr *AmaltheaSession) AdoptedSecrets() v1.SecretList {
...
-	if as.Spec.SessionLocation == Remote {
-		return []v1.PersistentVolumeClaim{}, []v1.Volume{}, []v1.VolumeMount{}
+	switch as.Spec.SessionLocation {
+	case Remote:
+		return as.RemoteSessionDataSources()
+	case Local:
+		return as.LocalSessionDataSources()
+	default:
+		panic("invalid session location")
+	}
```
leafty: We should do a little bit of defensive programming here: this method should work even if the CRD is outdated and doesn't have the location field.

See also: https://github.com/SwissDataScienceCenter/amalthea/pull/1121, https://github.com/SwissDataScienceCenter/amalthea/pull/1120

```suggestion
	if as.Spec.SessionLocation == Remote {
		return as.RemoteSessionDataSources()
	}
	// as.Spec.SessionLocation == Local
	return as.LocalSessionDataSources()
```

api/v1alpha1/amaltheasession_children.go
```diff
@@ -642,11 +645,92 @@ func (cr *AmaltheaSession) AdoptedSecrets() v1.SecretList {
...
+}
+
+func (as *AmaltheaSession) RemoteSessionDataSources() ([]v1.PersistentVolumeClaim, []v1.Volume, []v1.VolumeMount) {
+	pvcs := []v1.PersistentVolumeClaim{}
+	vols := []v1.Volume{
+		{
+			Name: fmt.Sprintf("%s-%s", prefix, as.Name),
+			VolumeSource: v1.VolumeSource{
+				Secret: &v1.SecretVolumeSource{
```
leafty, line 661: We should add `Optional:   ptr.To(false)`, no?

api/v1alpha1/amaltheasession_children.go
```diff
@@ -642,11 +645,92 @@ func (cr *AmaltheaSession) AdoptedSecrets() v1.SecretList {
...
+					SecretName:  as.InternalSecretName(),
+					DefaultMode: ptr.To(int32(256)), // decimal value of 0400 for the access flags (chmod-like)
```
```suggestion
					DefaultMode: ptr.To(int32(0400)),  // chmod: r-- --- ---
```

api/v1alpha1/amaltheasession_children.go
```diff
@@ -642,11 +645,92 @@ func (cr *AmaltheaSession) AdoptedSecrets() v1.SecretList {
...
+				},
+			},
+		},
+	}
+	volMounts := []v1.VolumeMount{
+		{
+			Name:      fmt.Sprintf("%s-%s", prefix, as.Name),
```
leafty, line 670: You will have two dashes

```suggestion
			Name:      fmt.Sprintf("%s%s", prefix, as.Name),
```

internal/common/data_connector.go
```diff
@@ -0,0 +1,198 @@
...
+	"path"
+	"strings"
+
+	"github.com/fernet/fernet-go"
+	"gopkg.in/ini.v1"
+)
+
+const UserSecretProxyFolder = "/secrets-user"
+const DataConnectorProxyFolder = "/secrets-dcs"
+const DataConnectorSecretProxyFolder = "/secrets-dcs-secrets"
```
leafty: Proxy?

internal/common/data_connector.go
```diff
@@ -0,0 +1,198 @@
...
+
+type DataConnector struct {
+	root string
```
leafty: Why is `root` a private field?

internal/common/data_connector.go
```diff
@@ -0,0 +1,198 @@
...
+	if dirEntries, err = os.ReadDir(dataConnectorSecretMountPoint); err != nil && !os.IsNotExist(err) {
+		return nil, err
+	}
+
+	decodedSecrets := make(map[string][]byte)
+	for _, dir := range dirEntries {
+		// TODO: Add support for hierarchies ?
+		if dir.IsDir() || strings.HasPrefix(dir.Name(), "..") {
+			continue
+		}
```
leafty, line 61: `os.ReadDir()` will not return `.` or `..`

```suggestion
		if dir.IsDir() {
			continue
		}
```

internal/common/data_connector.go
```diff
@@ -0,0 +1,198 @@
...
+	RemotePath string
+	MountOpt   string
+	VfsOpt     string
+
+	ExtraArgs []string
+}
+
+func (dc *DataConnector) fernetKey() (*fernet.Key, error) {
+	// the fernet key is mounted as part of the data source secret
+	if encodedKey, err := os.ReadFile(path.Join(dc.root, dc.Name, "secretKey")); err == nil {
```
leafty, line 34: Since this is a local file, you should use `filepath.Join()` and not `path.Join()`

internal/common/data_connector.go
```diff
@@ -0,0 +1,198 @@
...
+	if err != nil && !os.IsNotExist(err) {
+		return nil, err
+	}
+
+	// Override values with secrets
+	for k, v := range dataConnectorSecrets {
+		switch k {
+		case "pass":
+			// Put it in a file, which will allow for passing to `rclone obscure -` before being placed in the config file.
+			configFiles[k] = v
```
leafty, line 98: Why do we do this?

sambuc: The version of rclone we use on the remote nodes requires the sensitive values to be obscured in the config file.

Obscuring on our side would require duplicating and/or importing rclone libraries, and then to ensure the version of the libraries is compatible with the version we install on the remote hosts.

In order to keep maintenance simple in the long run, I use the command provided with the version we install, so it stays in sync if the rclone project decides to change the algorithm / procedure they use for their obscuration process.

internal/common/data_connector.go
sambuc: Correct, but folders mounted from K8s secrets have hidden entries with a `..` prefix, like `..data` (I don't remember the exact name now) which are causing issues if I don't filter them out.

api/v1alpha1/amaltheasession_children.go
sambuc: Applied everywhere.

internal/common/data_connector.go
sambuc: I have renamed these constants with the previously agreed patterns.

Thanks for catching this.

internal/common/data_connector.go
sambuc: I have been taught to encapsulate and abstract. By default, as long as I don't **need** something to be public or exported, I keep it private.

This gives me or the next developer more margin to refactor or change things without breaking the visible / public API.

leafty: You were taught object oriented programming which go is not. Using unexported fields on a struct has adverse consequences: users outside the go package cannot construct the object (and must use a constructor method), copying the object will also copy the unexported fields, etc.

Given that DataConnector is purely a data holding type, all its fields should be exported.

leafty:
```suggestion
	Root string
```

Note: change required

internal/common/data_connector.go
sambuc: Thanks, I have reworked everywhere this was need in this file.

api/v1alpha1/amaltheasession_children.go
sambuc: done

api/v1alpha1/amaltheasession_children.go
sambuc: done

leafty (approved):
LGTM 👏

PR #1174, opened 2026-07-16
feat: Mount DC in HPC sessions

sambuc:
### PR Stack
* #1173
* 🐸 #1174

### DIFF
--- a/internal/remote/firecrest/session_script.sh
+++ b/internal/remote/firecrest/session_script.sh
@@ -36,10 +36,11 @@ esac
 
 : ${SESSION_DIR:="${PWD}"}
 : ${SESSION_WORK_DIR:="${SESSION_DIR}/work"}
-: ${SECRETS_DIR:="${SESSION_DIR}/secrets"}
+: ${SECRETS_DIR:="$(mktemp -d)"}
 : ${SECRETS_USER_DIR:="${SECRETS_DIR}/user/"}
 : ${SECRETS_DATA_CONNECTORS_DIR:="${SECRETS_DIR}/data_connectors"}
 : ${LOGS_DIR:="${SESSION_DIR}/logs"}
+: ${CACHE_DIR:="${SECRETS_DIR}/cache"}
 
 # Setup session environment
 export RENKU_MOUNT_DIR="${SESSION_WORK_DIR}"
@@ -47,8 +48,36 @@ export RENKU_WORKING_DIR="${SESSION_WORK_DIR}"
 # Force the frontend to listen on 127.0.0.1
 export RENKU_SESSION_IP="127.0.0.1"
 
+# Do not leave secrets on a shared fs, move it to the node where the session runs.
+mkdir -p "${SECRETS_DIR}"
+chmod 700 "${SECRETS_DIR}"
+mv "${SESSION_DIR}/secrets"/* "${SECRETS_DIR}"
+rmdir "${SESSION_DIR}/secrets"
+
 # Load the wstunnel secret
 export WSTUNNEL_SECRET="$(cat "${SECRETS_DIR}/wstunnel_secret")"
+
+# CamelCase to kebab-case conversion
+#
+# Usage:
+#     kebab_cased="$(to_kebab_case "[cC]amelCaseInput")"
+to_kebab_case() {
+    echo "${1:?"to_kebab_case: input string missing"}" | sed 's/\([A-Z]\)/-\1/g' | tr '[:upper:]' '[:lower:]'
+}
+
+# Convert camelCased key - value pairs stored in a flat json struct to command line arguments.
+#
+# WARNING: Whitespace and json reserved characters are stripped, so they cannot appear in keys nor values.
+#
+# Usage:
+#     arguments_string="$(to_rclone_mount_arguments "file_path" ["argument_prefix"])"
+to_rclone_mount_arguments() {
+    local filename=${1:?"to_rclone_mount_arguments: input file missing"}
+    local prefix=${2}
+    cat "${filename}" | tr -d '{} \t\r\n' | sed -e 's/","/"\n"/g' -e 's/":"/": "/g' | while IFS=': ' read -r key value; do
+        printf "%s%s=%s " "${prefix}" "$(to_kebab_case "${key}")" "${value}"
+    done
+}
 
 # Installs rclone
 #
@@ -135,8 +164,6 @@ for d in \
     SESSION_DIR \
     SESSION_WORK_DIR \
     SECRETS_DIR \
-    SECRETS_USER_DIR \
-    SECRETS_DATA_CONNECTORS_DIR \
     LOGS_DIR
 do
     echo "${d}: ${!d}"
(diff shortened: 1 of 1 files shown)


internal/remote/firecrest/session_script.sh
```diff
@@ -36,20 +36,49 @@ esac
...
 export RENKU_WORKING_DIR="${SESSION_WORK_DIR}"
 # Force the frontend to listen on 127.0.0.1
 export RENKU_SESSION_IP="127.0.0.1"
 
+# Do not leave secrets on a shared fs, move it to the node where the session runs.
+mkdir -p "${SECRETS_DIR}"
+chmod 700 "${SECRETS_DIR}"
+mv "${SESSION_DIR}/secrets"/* "${SECRETS_DIR}"
+rmdir "${SESSION_DIR}/secrets"
+
```
leafty, line 56: Why do we do this? Official CSCS docs have example where user credentials are stored in the HOME folder. I really do not see what we gain by doing this.

Example: https://docs.cscs.ch/software/container-engine/run/#third-party-and-private-registries

Having the credentials in the HOME folder allows user to re-access the data connectors even outside of a remote Renku session (e.g. connecting directly to Alps with SSH).

internal/remote/firecrest/session_script.sh
```diff
@@ -172,20 +199,66 @@ fi
...
+            mkdir -p "${mount_point}"
+            mkdir -p "${CACHE_DIR}/${n}"
+
+            # Make sure there is no stale mount, and if so clean it up.
+            fusermount3 -uz "${mount_point}" 2>/dev/null || true
+            sleep 1 # Let the state stabilise before mounting something there
+
+            # We do our best to make sure we do not leave around bad rclone
+            # state, but sometimes there is still something lingering on...
+            # so we have to add --allow-non-empty.
```
leafty, line 247: Why not do `rm -rf <mount_point>; mkdir -p <mount_point>`?

sambuc: I was using the `rm` / `mkdir` sequence, but it still failed occasionally.

I _think_ this is a (timing?) issue with the distributed FS.

I have not noticed any problems with the `--allow-non-empty` enabled, and the behaviour is stable.

internal/remote/firecrest/session_script.sh
sambuc: The official CSCS documentation assumes the **users stores themselves their credentials**, this means they know they are stored and where.

We are in a different use-case as this would be an **implicit** side-effect of using renku. I am not convinced most users would expect to have their credentials transferred and stored permanently on a system just because they started once a session on that specific cluster.

I think it is much more reasonable to have the users set their credentials by themselves when they need it. This solves the problem of the user **knowing** and be **aware** their credentials have been saved, and where, on that cluster.

On a more general note, when handling user credentials, I have been taught to restrict their presence to the strict minimum, both in terms of time and access:
  - leave the credentials only as long as required on a system for the action triggered by the user.
  - put them in a place with as little as possible access, through file permissions, and a placement as close as possible to the place of usage.

Finally, putting them in a system temporary folder also guarantees that they will be cleaned up at the latest when the machine shutdowns or reboots, if all our cleaning procedures were to fail for any reason.

internal/remote/firecrest/session_script.sh
```diff
@@ -36,20 +36,49 @@ esac
...
 # Setup session environment
 export RENKU_MOUNT_DIR="${SESSION_WORK_DIR}"
 export RENKU_WORKING_DIR="${SESSION_WORK_DIR}"
 # Force the frontend to listen on 127.0.0.1
 export RENKU_SESSION_IP="127.0.0.1"
 
+# Do not leave secrets on a shared fs, move it to the node where the session runs.
+mkdir -p "${SECRETS_DIR}"
+chmod 700 "${SECRETS_DIR}"
+mv "${SESSION_DIR}/secrets"/* "${SECRETS_DIR}"
```
leafty: This `mv` command will not work for secret files starting with a dot `.`

internal/remote/firecrest/session_script.sh
leafty: This is all nice if everything goes well, but the secrets will stay in $HOME if there is something that prevents the session from starting, for example:
1. User starts a CSCS session
2. The job is put in the slurm queue, the job is pending for a long time or is rejected (could be a maintenance day)
3. The user shuts down the Renku session

In this case, you never start the `session_script.sh` so the secrets will stay in $HOME.

internal/remote/firecrest/session_script.sh
```diff
@@ -172,20 +199,66 @@ fi
 srun_param_container_image="--container-image ${REMOTE_SESSION_IMAGE}"
 srun_param_workdir="--container-workdir ${SESSION_WORK_DIR}"
 srun_param_mounts=#{{SESSION_MOUNTS_PLACEHOLDER}}
+# We cannot generate directly the local secrets path from the proxy as the final path is only known after ${SECRETS_DIR} as been set.
+srun_param_mounts=$(echo ${srun_param_mounts} | sed -e "s,${SESSION_DIR}/secrets,${SECRETS_DIR},g")
+
+# Mount Data Connectors, if any
+if [ -d  "${SECRETS_DATA_CONNECTORS_DIR}" ]; then
+    (# Run in a sub shell to scope the temporary variables
+        for dc in "${SECRETS_DATA_CONNECTORS_DIR}"/*; do
```
leafty, line 208: Brittle code: what if a data connector name starts with a dot? What if there is some random folder? Maybe you should push the list of DCs as an env var so that we do not rely on `ls`.

sambuc: This directory contains folders with names generated by our code, in the format `${SESSION_PREFIX}-${HASH}-ds-${sequence_number}`, where `sequence_number` is an integer starting from 0.

These contains the uploaded config values. Those are not the user-configured mount points.

internal/remote/firecrest/session_script.sh
sambuc: Thanks for the catch, I am fixing this.

internal/remote/firecrest/session_script.sh
leafty: This may still lead to issues if the user adds and deletes DCs between session starts.

sambuc: The `${SECRETS_DATA_CONNECTORS_DIR}` is removed and recreated on each session stop & restart, to avoid issues with stale credentials _and_ configurations.

internal/remote/firecrest/session_script.sh
sambuc: Fixed by using a glob pattern which matches everything except `.` and `..`

internal/remote/firecrest/session_script.sh
sambuc: Just as a reminder, all the uploads from the proxy container are within `SESSION_DIR`, which is on `/scratch/[..]/`. 

If the session script does not execute, they will never be moved anywhere outside of `/scratch`.

leafty: I mean ok, scratch and not home, but the credentials would still be lingering there...

internal/remote/firecrest/session_script.sh
```diff
@@ -36,20 +36,49 @@ esac
...
 # Setup session environment
 export RENKU_MOUNT_DIR="${SESSION_WORK_DIR}"
 export RENKU_WORKING_DIR="${SESSION_WORK_DIR}"
 # Force the frontend to listen on 127.0.0.1
 export RENKU_SESSION_IP="127.0.0.1"
 
+# Do not leave secrets on a shared fs, move it to the node where the session runs.
+mkdir -p "${SECRETS_DIR}"
+chmod 700 "${SECRETS_DIR}"
+mv "${SESSION_DIR}/secrets"/{.[!.]*,*} "${SECRETS_DIR}"
```
leafty, line 54: Wouldn't this be simpler if you moved the `${SESSION_DIR}/secrets` folder itself? Because now, you would leave out files like `..my_secret.txt`

PR #1185, opened 2026-07-21
feat: adapt requeuing when session is hibernated

sgaist:
## Describe your changes

Currently, the operator checks on all sessions every 10 seconds.

This is not needed anymore once a pod has been hibernated.
Therefore queue for the MaxHibernationDuration.

If it's awaken before, there will be an event that will trigger an earlier reconciliation.


leafty (changes requested):
This change prevents prometheus metrics to be collected for a paused session.

See: https://github.com/SwissDataScienceCenter/amalthea/blob/9afb5383f45d73816c0df74d8e689525f7b19c31/internal/controller/amaltheasession_controller.go#L258

sgaist:
Thinking about this a bit more:
- Can we consider bringing the interval to say 60s ? I don't think there's a need for a lower resolution check for hibernated sessions
- Should we consider using a different mechanism to log hibernated sessions statistics ?

olevski:
> Should we consider using a different mechanism to log hibernated sessions statistics ?

I agree, yes. We should not be prevented from adjusting this just for the sake of publishing metrics.

> Can we consider bringing the interval to say 60s ? I don't think there's a need for a lower resolution check for hibernated sessions

So in the meantime we can go to 60 sec for hibernated sessions. I would go for merging this with 60 sec hibernation interval and without breaking the metrics. 

Then in a followup pr lets change the metrics collection and further increase the requeue interval as you originally intended. Wdyt? If not we can do everything all at once. Let me know @sgaist

sgaist:
Sounds good to me

olevski:
@sgaist the cyclomatic complexity on the linter is triggering. You can ignore that rule on the current fuction or you can just fully disable that. It is one of the default checks on the linter we got for "free" but we never consciously turned it on.

sgaist:
@olevski I added some cleanups as they where hanging things that I wanted to tackle anyway and it fixed the cyclomatic complexity issue.

leafty (approved):

PR #1205, opened 2026-08-06
feat: propagate session resource requests to SLURM for FirecREST remote sessions

SalimKayal:
## Summary

For remote sessions dispatched via FirecREST, resource requests defined in `AmaltheaSession.Spec.Session.Resources` are now forwarded to the SLURM job script as `#SBATCH` directives. An opt-out environment variable preserves the previous behavior for sites that do not want resource-class values passed to the grid.

## Motivation and context

FirecREST remote sessions were already using SLURM to schedule the session container, but the CPU, memory, and GPU values configured in the custom resource were ignored by the grid. This change lets the grid honor the resource class the user requested, while still allowing sites to fall back to grid defaults when a value is omitted.

## Behavior

- `RSC_SESSION_CPU` is emitted as `#SBATCH --cpus-per-task=<value>`.
- `RSC_SESSION_MEMORY` is emitted as `#SBATCH --mem=<value>M`.
- `RSC_SESSION_GPUS` is emitted as `#SBATCH --gpus=<value>`.
- If any env var is empty, its directive is omitted so the grid default applies.
- If `RSC_FIRECREST_IGNORE_RESOURCE_CLASS_VALUES` is set to the literal string `"true"`, all resource directives are omitted and the previous behavior is preserved.

## Changes

- `api/v1alpha1/amaltheasession_children.go`
  - `sessionContainerRemote()` now exposes three new environment variables derived from `session.Resources`:
    - `RSC_SESSION_CPU` — integer CPU count, computed as `ceil(milliCPU / 1000)`.
    - `RSC_SESSION_MEMORY` — memory in mebibytes.
    - `RSC_SESSION_GPUS` — sum of all resource keys ending in `/gpu` (supports multi-vendor GPU requests).
  - Falls back from `Requests` to `Limits` for CPU and memory when requests are empty.

- `internal/remote/firecrest/controller.go`
  - `addSbatchDirectivesToScript()` consumes the new env vars and emits the directives described in **Behavior**.
  - Existing directives (`--nodes=1`, `--ntasks-per-node=1`, `--partition`, `--account`) remain unchanged.

- Tests
  - Added `TestSessionContainerRemoteResources` and `TestSessionContainerRemoteResourcesLimitsFallback` in `api/v1alpha1/amaltheasession_children_test.go`.
  - Extended `TestRenderSessionScriptStatic` in `internal/remote/firecrest/controller_test.go` with subtests for resources provided, ignore-flag omission, and CPU-only resources.

### DIFF
--- a/api/v1alpha1/amaltheasession_children.go
+++ b/api/v1alpha1/amaltheasession_children.go
@@ -1052,33 +1052,14 @@ func (cr *AmaltheaSession) sessionContainerRemote(volumeMounts []v1.VolumeMount)
 
 	var cpuValue, memoryValue, gpuValue string
 
-	if cpuQ, ok := resources.Requests[v1.ResourceCPU]; ok && !cpuQ.IsZero() {
-		cpuValue = strconv.FormatInt((cpuQ.MilliValue()+999)/1000, 10)
-	} else if cpuQ, ok := resources.Limits[v1.ResourceCPU]; ok && !cpuQ.IsZero() {
-		cpuValue = strconv.FormatInt((cpuQ.MilliValue()+999)/1000, 10)
+	if q := resourceValue(resources, v1.ResourceCPU); !q.IsZero() {
+		cpuValue = strconv.FormatInt((q.MilliValue()+999)/1000, 10) // ceil(milli/1000)
 	}
-
-	if memQ, ok := resources.Requests[v1.ResourceMemory]; ok && !memQ.IsZero() {
-		memoryValue = strconv.FormatInt(memQ.Value()/(1024*1024), 10)
-	} else if memQ, ok := resources.Limits[v1.ResourceMemory]; ok && !memQ.IsZero() {
-		memoryValue = strconv.FormatInt(memQ.Value()/(1024*1024), 10)
+	if q := resourceValue(resources, v1.ResourceMemory); !q.IsZero() {
+		memoryValue = strconv.FormatInt(q.Value()/(1024*1024), 10)
 	}
-
-	var gpuCount int64
-	for name, q := range resources.Requests {
-		if strings.HasSuffix(string(name), "/gpu") {
-			gpuCount += q.Value()
-		}
-	}
-	if gpuCount == 0 {
-		for name, q := range resources.Limits {
-			if strings.HasSuffix(string(name), "/gpu") {
-				gpuCount += q.Value()
-			}
-		}
-	}
-	if gpuCount > 0 {
-		gpuValue = strconv.FormatInt(gpuCount, 10)
+	if q := resourceValue(resources, v1.ResourceName("nvidia.com/gpu")); !q.IsZero() {
+		gpuValue = strconv.FormatInt(q.Value(), 10)
 	}
 
 	sessionContainer.Env = append(
@@ -1120,6 +1101,18 @@ func (cr *AmaltheaSession) sessionContainerRemote(volumeMounts []v1.VolumeMount)
 	}
 
 	return sessionContainer
+}
+
+// resourceValue returns q[name] from Requests, falling back to Limits
+// (what the node actually reserves), or zero when neither is set.
+func resourceValue(res v1.ResourceRequirements, name v1.ResourceName) resource.Quantity {
+	if q, ok := res.Requests[name]; ok && !q.IsZero() {
+		return q
+	}
+	if q, ok := res.Limits[name]; ok && !q.IsZero() {
+		return q
+	}
+	return resource.Quantity{}
 }
 
 // tunnelContainer returns the tunnel container for remote sessions
--- a/api/v1alpha1/amaltheasession_children.go
+++ b/api/v1alpha1/amaltheasession_children.go
@@ -11,6 +11,7 @@ import (
 	"net/url"
 	"os"
 	"sort"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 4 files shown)


SalimKayal:
Question: I realize I set the default to set the resource variables form the custom resource, perhaps it would be saner to default to the current behavior for backward compatibility instead ?

I did it like this because to me it makes more sense to but it could break stuff, perhaps the other way around is better ?

api/v1alpha1/amaltheasession_children.go
```diff
@@ -1033,6 +1034,46 @@ func (cr *AmaltheaSession) sessionContainerRemote(volumeMounts []v1.VolumeMount)
 		},
 	)
 
+	resources := session.Resources
+
+	var cpuValue, memoryValue, gpuValue string
```
leafty, line 1055: The logic seems a bit convoluted to me.

```go
// CPU handling
var cpuValue string
cpuQ := resource.Quantity{}
if q, ok := resources.Requests[v1.ResourceCPU]; ok && q.Sign() > 0 {
	cpuQ = q
} else if q, ok := resources.Limits[v1.ResourceCPU]; ok && q.Sign() > 0 {
	cpuQ = q
}
if !cpuQ.IsZero() {
	cpuValue = strconv.FormatInt(cpuQ.Value(), 10)
}
```

```go
// Memory handling
var memValue string
memQ := resource.Quantity{}
if q, ok := resources.Requests[v1.ResourceMemory]; ok && q.Sign() > 0 {
	memQ = q
} else if q, ok := resources.Limits[v1.ResourceMemory]; ok && q.Sign() > 0 {
	memQ = q
}
if !memQ.IsZero() {
	memValue = strconv.FormatInt(memQ.Value()/1<<20, 10)
}
```

api/v1alpha1/amaltheasession_children_test.go
```diff
@@ -64,3 +65,63 @@ func TestGetPodEventsSorted(t *testing.T) {
 	assert.Nil(t, err)
 	assert.Equal(t, res.Items, []v1.Event{ev3, ev2, ev1})
 }
+
+func TestSessionContainerRemoteResources(t *testing.T) {
```
leafty, line 69: Note: should have a test with cpu and/or mem not set.

internal/remote/firecrest/controller_test.go
```diff
@@ -58,6 +58,38 @@ func TestRenderSessionScriptStatic(t *testing.T) {
...
+		t.Setenv("RSC_FIRECREST_IGNORE_RESOURCE_CLASS_VALUES", "true")
+
+		script := renderSessionScriptStatic(sessionScript, partition, &fileSystems, secretsPath)
+		assert.NotContains(t, script, "--cpus-per-task")
+		assert.NotContains(t, script, "--mem")
+		assert.NotContains(t, script, "--gpus")
+	})
+
+	t.Run("only cpu provided", func(t *testing.T) {
+		t.Setenv("RSC_SESSION_CPU", "4")
```
leafty, line 86: Note: the current logic will have `t.Setenv("RSC_SESSION_MEMORY", "")` and `t.Setenv("RSC_SESSION_GPUS", "")`.

This is because the env vars are always emitted with the code in api/v1alpha1/amaltheasession_children.go

api/v1alpha1/amaltheasession_children.go
SalimKayal: refactored.

api/v1alpha1/amaltheasession_children_test.go
SalimKayal: done

internal/remote/firecrest/controller_test.go
SalimKayal: fixed

api/v1alpha1/amaltheasession_children.go
```diff
@@ -1038,33 +1038,14 @@ func (cr *AmaltheaSession) sessionContainerRemote(volumeMounts []v1.VolumeMount)
 
 	var cpuValue, memoryValue, gpuValue string
 
-	if cpuQ, ok := resources.Requests[v1.ResourceCPU]; ok && !cpuQ.IsZero() {
-		cpuValue = strconv.FormatInt((cpuQ.MilliValue()+999)/1000, 10)
-	} else if cpuQ, ok := resources.Limits[v1.ResourceCPU]; ok && !cpuQ.IsZero() {
-		cpuValue = strconv.FormatInt((cpuQ.MilliValue()+999)/1000, 10)
+	if q := resourceValue(resources, v1.ResourceCPU); !q.IsZero() {
+		cpuValue = strconv.FormatInt((q.MilliValue()+999)/1000, 10) // ceil(milli/1000)
```
leafty: What is this `+999`?

SalimKayal: basically `strconv.FormatInt` will take a floor of the value so by adding `999` to `q.MilliValue()` we ensure we have a ceil.

leafty: I don't understand why we don't do `q.Value()` as it will already return a `ceil()`.

api/v1alpha1/amaltheasession_children.go
```diff
@@ -1033,6 +1034,27 @@ func (cr *AmaltheaSession) sessionContainerRemote(volumeMounts []v1.VolumeMount)
 		},
 	)
 
+	resources := session.Resources
+
+	var cpuValue, memoryValue, gpuValue string
+
+	if q := resourceValue(resources, v1.ResourceCPU); !q.IsZero() {
+		cpuValue = strconv.FormatInt((q.MilliValue()+999)/1000, 10) // ceil(milli/1000)
```
```suggestion
		cpuValue = strconv.FormatInt((q.Value(), 10)
```

api/v1alpha1/amaltheasession_children.go
```diff
@@ -1033,6 +1034,27 @@ func (cr *AmaltheaSession) sessionContainerRemote(volumeMounts []v1.VolumeMount)
...
+	}
+	if q := resourceValue(resources, v1.ResourceMemory); !q.IsZero() {
+		memoryValue = strconv.FormatInt(q.Value()/(1024*1024), 10)
```
leafty: Note: maybe define a const `oneMebiByte`

api/v1alpha1/amaltheasession_children.go
SalimKayal: Oh THANKS. I learned something.

api/v1alpha1/amaltheasession_children.go
SalimKayal: done

leafty:
> Question: I realize I set the default to set the resource variables form the custom resource, perhaps it would be saner to default to the current behavior for backward compatibility instead ?
> 
> I did it like this because to me it makes more sense to but it could break stuff, perhaps the other way around is better ?

Yes, it may be better to flip the logic for passing the values to SLURM.

internal/remote/firecrest/controller.go
```diff
@@ -757,6 +757,19 @@ func addSbatchDirectivesToScript(sessionScript, partition string) string {
 	if slurmAccount != "" {
 		directives = append(directives, fmt.Sprintf("#SBATCH --account=%s", slurmAccount))
 	}
+
+	if os.Getenv("RSC_FIRECREST_IGNORE_RESOURCE_CLASS_VALUES") != "true" {
```
leafty: Maybe we should flip the logic:


```suggestion
	if strings.ToLower(os.Getenv("RSC_FIRECREST_FORWARD_RESOURCE_VALUES")) == "true" {
```

SalimKayal: I still have to change this in data-services.

leafty (approved):
LGTM

PR #1230, opened 2026-09-10
feat: open port 2222 on amalthea session service

SalimKayal:


api/v1alpha1/amaltheasession_children.go
```diff
@@ -49,6 +50,11 @@ const AuthProxyMetaPort int32 = 65534
 const secondProxyPort int32 = 65533
 const RemoteSessionControllerPort int32 = 65532
 const TunnelPort int32 = 65531
+const SSHPort int32 = 2222
```
leafty: 2222 is not reserved, so it could be in conflict with the session.

